PR #10, opened 2023-09-06
fix oauth path being lost

constantins2001:
closes https://github.com/Samedis-care/samedis-care-issues/issues/1166


SCjona (approved):

neongrau:
undo now

PR #23, opened 2023-10-17
fix wrong endpoint

neongrau:
as the endpoint requires app specific tenant candos it must be below /apps for the correct cando determination

### DIFF
--- a/config/routes/v1.rb
+++ b/config/routes/v1.rb
@@ -12,6 +12,7 @@ namespace :v1 do
     end
     resources :tenant, module: :tenant, only: [] do
       resources :groups, only: [:index]
+      resources :users, only: [:index, :show, :update, :destroy]
     end
   end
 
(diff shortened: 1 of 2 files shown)


neongrau:
@SCjona requires frontend change for used endpoint

`DELETE /api/v1/access_control/tenant/:tenant_id/users/:user_id`

must be changed to

`DELETE /api/v1/access_control/apps/:app_id/tenant/:tenant_id/users/:user_id`

config/routes/v1.rb
```diff
@@ -6,10 +6,12 @@
     end
     resources :apps, module: :apps, only: [] do
       resources :groups, only: [:index]
+      resources :tenant, module: :tenant, only: [] do
+        resources :users, only: [:index, :show, :update, :destroy]
+      end
     end
     resources :tenant, module: :tenant, only: [] do
       resources :groups, only: [:index]
-      resources :users, only: [:index, :show, :update, :destroy]
```
SCjona, line 12: you removed the route, but what about the controller?

SCjona (approved):

PR #135, opened 2025-02-18
ensures default standard user group exists

neongrau:
if defined per app, children of tenant_profiles will be created as groups and ensured on first usage of profiles_ou


neongrau:
app-config repo has the group as children and needs to be up to date before seeding

SCjona (approved):

PR #261, opened 2026-06-30
APAC migration: leak guard + verify correct for a live target

neongrau:
Follow-up to #260. Once the APAC cluster is live and accepting native signups/logins, it legitimately accumulates records that are **not** part of the migration set. The previous whitelist-based safety checks ("everything on the target must be in the APAC set") then false-alarmed on that native data — both `apac:copy`'s leak guard and `apac:verify` aborted on records created by test logins / `db:seed` on the live APAC cluster.

This switches both checks to the correct invariant.

## Changes
- **Leak guard** now flags by an **EU-source-tenant blacklist** (`tenant_id ∈ source EU tenants`) instead of the APAC whitelist. Records created natively on the live APAC cluster (not present in the source) are correctly ignored; genuine EU leaks still fail loudly. Covers mappings, tenants, invites and cached tenant refs on users.
- **`tenant_set_resolver`** gains `eu_tenant_ids` (all source tenants minus the APAC set) backing the blacklist.
- **`apac:verify`** checks **completeness** (every source `_id` present on the target) instead of exact count equality. Extra native target docs are reported as `(+N native on target)`; the task fails only when source docs are actually **missing**.
- **Spec**: the fake Mongo client now evaluates `$in`, so the EU-vs-native distinction is genuinely exercised.

## Verification
- `bundle exec rspec spec/lib/apac_migration` → 13 examples, 0 failures
- rubocop clean (only the intentional `puts`/`abort` in the rake/CLI tooling remain, consistent with `lib/tasks/app.rake`)
- On the live APAC target: `apac:verify` now reports `OK (+58 native on target)` and `Leak guard passed` instead of aborting.

### DIFF
--- a/lib/apac_migration/leak_guard.rb
+++ b/lib/apac_migration/leak_guard.rb
@@ -8,41 +8,36 @@ module ApacMigration
     end
 
     def assert!(target_client)
-      apac = @resolver.apac_tenant_ids
-      # Invite#tenant_id is `type: Object` and stored as a mix of String and
-      # BSON::ObjectId, so match both representations (Mapping#tenant_id is a
-      # clean ObjectId and only needs `apac`).
-      apac_mixed = apac + apac.map(&:to_s)
+      # Blacklist of EU tenants that exist in the SOURCE. Anything on the target
+      # tied to one of these is a genuine leak; records created natively on the
+      # (live) APAC cluster are not in this set and are correctly ignored.
+      eu = @resolver.eu_tenant_ids
+      # Invite#tenant_id is `type: Object`, stored as a mix of String and
+      # BSON::ObjectId — match both. Mapping#tenant_id is a clean ObjectId.
+      eu_mixed = eu + eu.map(&:to_s)
       errors = []
 
-      # 1. Every Mapping on the target must point at an APAC tenant.
+      # 1. No Mapping on the target may point at an EU source tenant.
       bad_map = target_client['actors'].find(
         '_type' => 'Actors::Mapping',
-        'tenant_id' => { '$nin' => apac }
+        'tenant_id' => { '$in' => eu }
       ).first
-      errors << "actors: Mapping #{bad_map['_id']} has non-APAC tenant_id #{bad_map['tenant_id']}" if bad_map
+      errors << "actors: Mapping #{bad_map['_id']} points at EU tenant #{bad_map['tenant_id']}" if bad_map
 
-      # 1b. Every Tenant actor on the target must be in the APAC set (guards the
-      #     structural-ancestor copy from ever pulling in a sibling tenant).
+      # 2. No EU source Tenant may exist on the target.
       bad_tenant = target_client['actors'].find(
         '_type' => 'Actors::Tenant',
-        '_id' => { '$nin' => apac }
+        '_id' => { '$in' => eu }
       ).first
-      errors << "actors: Tenant #{bad_tenant['_id']} is not in the APAC set" if bad_tenant
+      errors << "actors: EU Tenant #{bad_tenant['_id']} present on target" if bad_tenant
 
-      # 2. Every Invite on the target must belong to an APAC tenant.
-      bad_inv = target_client['invites'].find('tenant_id' => { '$nin' => apac_mixed }).first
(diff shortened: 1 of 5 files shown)


lib/apac_migration/leak_guard.rb
```diff
@@ -8,41 +8,36 @@ def initialize(resolver)
...
-      bad_user = target_client['users'].find(
-        '$or' => [
-          { 'tenants_cached' => { '$nin' => [nil, []] } },
-          { 'tenant_candos_cached' => { '$ne' => nil } },
-          { 'tenant_access_group_ids' => { '$nin' => [nil, {}] } }
-        ]
-      ).first
-      errors << "users: #{bad_user['_id']} still has EU cache fields populated" if bad_user
+      # 4. No user on the target may reference an EU tenant in its cached set.
+      bad_user = target_client['users'].find('tenants_cached' => { '$in' => eu_mixed }).first
```
SCjona, line 44: The user-cache leak check was narrowed from "any EU cache field populated" to "`tenants_cached` contains an EU id" — but `tenant_candos_cached` and `tenant_access_group_ids` are now **not guarded at all**.

Both of those fields are **keyed by tenant_id** (see `app/models/user.rb:104-108` — `tenant_access_group_ids` is `{ tenant_id => [group_ids] }`, and `tenant_candos_cached` is built per-tenant in `actors/mapping.rb`). So an EU tenant reference can live as a **hash key** in either field, and the new `'tenants_cached' => { '$in' => eu_mixed }` query (which only matches array *elements*) will never see it.

The copy transform (`apac_user_cache_clear`) does null all three at copy time, so this is a backstop rather than the primary defense — but the old guard's whole value was catching the case where that clearing *didn't* happen. Dropping these two from the guard removes that safety net on a privacy-critical migration. The "non-empty" check false-alarmed on native users (the correct motivation to change it), but the EU-aware equivalent would be to test whether either hash has an EU tenant id as a key, e.g. an `$or` over `{ "tenant_access_group_ids.<eu_id>" => { '$exists' => true } }` / `{ "tenant_candos_cached.<eu_id>" => { '$exists' => true } }` per EU id. Worth restoring coverage for these two fields rather than only `tenants_cached`.

lib/tasks/apac_migration.rake
```diff
@@ -283,20 +283,26 @@ namespace :apac do
...
+    # on the target are native APAC records (e.g. live signups) — expected on a
+    # live cluster, so they are reported but not a failure.
+    complete = true
     checks.each do |label, (model, selector)|
-      src = model.collection.find(selector).count
-      tgt = target[model.collection.name.to_s].count_documents(selector)
-      match = src == tgt
-      ok &&= match
-      puts "  #{label.ljust(20)} source=#{src}  target=#{tgt}  #{match ? 'OK' : 'MISMATCH'}"
+      src_ids = model.collection.distinct('_id', selector)
```
SCjona, line 291: Minor / operational: switching from `count_documents` to `distinct('_id', …)` materializes the full id set in memory and is subject to MongoDB's 16MB BSON result cap on `distinct` (~1.3M ObjectIds for the `actors (subtree)` check). The APAC slice is presumably well under that, so this is likely fine in practice — but it's a hard ceiling the old count-based check didn't have. If a large tenant subtree is ever in scope this would error rather than report a mismatch; an aggregation-based `_id` diff (or chunking) would be more robust. Flagging for awareness, not blocking.

SCjona (commented):
The whitelist→blacklist switch is the right call for a live APAC target, and the reasoning in the comments/PR description is clear and convincing: native records (absent from the source) are correctly ignored, while genuine EU-source leaks still fail loudly. The spec change (a fake client that actually evaluates `$in`) genuinely exercises the EU-vs-native distinction. The `verify` switch from count-equality to `_id`-set completeness (`missing` fails, `extra` reported as native) is also correct for a live cluster.

Two things worth a look before merge — one is a real reduction in the leak guard's coverage on the Datenschutz-critical path, the other is operational. Neither breaks the core invariant.

Also note `docs/apac_migration.md:154-155` still describes the old whitelist behaviour ("no Mapping/Invite with a non-APAC tenant_id") — worth updating to the new blacklist semantics.

SCjona (approved):

PR #264, opened 2026-07-07
security: tenant-scope authorization across all tenant-nested endpoints

neongrau:
Closes the Broken Access Control gaps (OWASP A01) on every tenant-nested IM endpoint, where a per-tenant cando (`tenant.admin`, `access-control.*`, `list-tenant-users.*`) was checked against the caller's **global** candos (union across all tenants) while the target tenant came from `params[:tenant_id]`.

### Verified (local_dev)
Users holding **only** `tenant.admin` (no app-wide cando), members of their own tenant(s), were granted by `authorize()` on a **foreign** tenant (zero candos there), target `m.schiffner@sana-mtsz.de`:
```
authorize() ACTUAL (global_candos) => true    CORRECT per-tenant => false
```
`apps/tenants/users` permits `:set_password` → cross-tenant account takeover. 2,212 external users hold `tenant.admin`.

## Two fix patterns, per endpoint intent

**A) Own-tenant self-service → `tenant_authorize`** (cando checked against a tenant the user is a MEMBER of; `tenant.admin` still works, but only for the user's own tenant):
- `apps/tenants/users` (C-1 — the account-takeover path)
- `access_control/tenant/profiles` (H-6), `access_control/tenant/roles` (M-9) — matching sibling `access_control/tenant/{users,groups}`
- `picker/tenant/users` / ident_accounts
- `json_api_controller`: `tenant_authorize` now also enforces the OTP gate (closes a latent MFA gap)

**B) Internal-admin-only → drop `~/tenant.admin`, gate to app-wide `~/app-tenant.admin`** (global `authorize` intentional; these are not customer surfaces):
- `apps/tenants/organizations/actor_roles` (C-2 — privilege escalation)
- `apps/tenants/organizations/mappings` (C-3), `apps/tenants/users/actors` (C-3)
- `apps/tenants/organizations`, `apps/tenants/organizations_tree`
- `apps/tenants/users/roles`, `apps/tenants/users/functionalities`
- `apps/tenants/organizations/picker/mappable_users`, `apps/tenants/picker/mappable_users`, `apps/tenants/picker/user_organization`

Every touched controller carries an inline SECURITY / pen-test note explaining the intent, so the choice reads as deliberate at the next review. Regression guard spec covers the `tenant_authorize` set (8 examples, green).

## Residual (accepted / to verify on prod)
`app-tenant.admin` is a trusted app-wide admin cando and remains able to assign/administer across tenants — keep it restricted to trusted operators; verify holders on prod (in local_dev: internal + the maintainer's own accounts).

## Testing
- `rspec spec/controllers/api/v1/tenant_scoped_authorization_spec.rb` → 8 examples, 0 failures.
- Smoke: every changed controller loads (`ruby -c` clean); before-chains correct.
- ⚠️ Repo lacks request-spec factories/auth helpers; behavior evidenced by the local_dev PoC.

### DIFF
--- a/app/controllers/api/v1/apps/tenants/organizations_controller.rb
+++ b/app/controllers/api/v1/apps/tenants/organizations_controller.rb
@@ -41,13 +41,16 @@ class Api::V1::Apps::Tenants::OrganizationsController < Api::V1::JsonApiControll
     super+[:organization_id]
   end
 
+  # SECURITY (pen-test 2026-07): internal-admin-only. ~/tenant.admin intentionally
+  # excluded — it is a per-tenant customer cando; app-tenant.admin is app-wide by
+  # design, so the global authorize is correct here. Do not re-add ~/tenant.admin.
   def cando
     CANDO.merge({
-      show:    %w(~/app-tenant.admin ~/tenant.admin),
-      index:   %w(~/app-tenant.admin ~/tenant.admin),
-      create:  %w(~/app-tenant.admin ~/tenant.admin),
-      update:  %w(~/app-tenant.admin ~/tenant.admin),
-      destroy: %w(~/app-tenant.admin ~/tenant.admin)
+      show:    %w(~/app-tenant.admin),
+      index:   %w(~/app-tenant.admin),
+      create:  %w(~/app-tenant.admin),
+      update:  %w(~/app-tenant.admin),
+      destroy: %w(~/app-tenant.admin)
     })
   end
 
--- a/app/controllers/api/v1/access_control/tenant/profiles_controller.rb
+++ b/app/controllers/api/v1/access_control/tenant/profiles_controller.rb
@@ -18,6 +18,11 @@ class Api::V1::AccessControl::Tenant::ProfilesController < Api::V1::JsonApiContr
     action_suffix: 'Custom profiles'
   }.freeze
 
+  # Tenant-scoped authorization — own-tenant only, matching the sibling
+  # access_control/tenant/{users,groups} controllers (security audit H-6).
+  skip_before_action :authorize
+  before_action :tenant_authorize
+
   PERMIT_CREATE = [
     :title,
     :name,
--- a/app/controllers/api/v1/access_control/tenant/roles_controller.rb
+++ b/app/controllers/api/v1/access_control/tenant/roles_controller.rb
@@ -13,6 +13,11 @@ class Api::V1::AccessControl::Tenant::RolesController < Api::V1::JsonApiControll
 
   SWAGGER = { tag: 'Access Control', name: 'Tenant Role', header: 'Manage roles within tenant access control' }
 
+  # Tenant-scoped authorization — own-tenant only, matching the sibling
+  # access_control/tenant/{users,groups} controllers (security audit M-9).
+  skip_before_action :authorize
+  before_action :tenant_authorize
+
   undef_method :create
   undef_method :update
   undef_method :destroy
--- a/app/controllers/api/v1/apps/tenants/organizations/actor_roles_controller.rb
+++ b/app/controllers/api/v1/apps/tenants/organizations/actor_roles_controller.rb
@@ -12,7 +12,24 @@ class Api::V1::Apps::Tenants::Organizations::ActorRolesController < Api::V1::Jso
 
   PERMIT_CREATE = [:role_id]
 
-  SWAGGER = { tag: 'Tenant Org Roles', name: 'Organization Role', header: 'Manage roles within a tenant organization' }
+  SWAGGER = {
+    tag: 'Tenant Org Roles',
+    name: 'Organization Role',
+    header: 'Manage roles within a tenant organization'
+  }
+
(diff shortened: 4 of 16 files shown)


app/controllers/api/v1/apps/tenants/organizations_controller.rb
```diff
@@ -41,13 +41,16 @@ def json_api_permits
...
+  # excluded — it is a per-tenant customer cando; app-tenant.admin is app-wide by
+  # design, so the global authorize is correct here. Do not re-add ~/tenant.admin.
   def cando
     CANDO.merge({
-      show:    %w(~/app-tenant.admin ~/tenant.admin),
-      index:   %w(~/app-tenant.admin ~/tenant.admin),
-      create:  %w(~/app-tenant.admin ~/tenant.admin),
-      update:  %w(~/app-tenant.admin ~/tenant.admin),
-      destroy: %w(~/app-tenant.admin ~/tenant.admin)
+      show:    %w(~/app-tenant.admin),
```
SCjona, line 49: Pattern B behavior change worth an explicit confirmation before merge: dropping `~/tenant.admin` here (and on `organizations_tree`, `users/{roles,functionalities,actors}`, `organizations/{mappings,actor_roles}`, and the `picker/mappable_users` / `user_organization` endpoints) turns these into `app-tenant.admin`-only surfaces. That is exactly right *if* they are never used by an ordinary tenant admin in IMF — but org-structure management (`organizations` / `organizations_tree`) is the most plausible customer-facing tenant-admin workflow of the set. Since candos come from IM and cannot be verified from this repo, please confirm against IMF permission usage (readPermission/editPermission on these resources) that no legitimate `tenant.admin`-only user drives these — otherwise they will start getting 403s. The inline note documents intent well; this is just asking to validate the "not a customer surface" premise for the org endpoints specifically.

SCjona (commented):
Reviewed the tenant-scope authorization changes against the auth flow (`json_api_controller#tenant_authorization` vs `#authorization`, `cando_any?`, `current_tenant`) and the routing for every touched endpoint. The security reasoning is sound and the implementation is correct — this is a solid fix for the broken-access-control gap.

**Verified:**
- `tenant_authorization` → `cando_any?(reqs, current_tenant)` evaluates candos against `current_tenant`, which is `current_user.tenants.find { |t| t[:id] == params[:tenant_id] }` — i.e. only tenants the caller is a *member* of. This genuinely closes the "global candos union" bypass that `authorization`/`current_user.cando_any?` exhibited.
- Every Pattern A controller receives `tenant_id` as a **path param** (`/access_control/tenant/:tenant_id/...`, `/apps/:app_id/tenants/:tenant_id/users`, `/picker/tenant/:tenant_id/users`), so `current_tenant` resolves correctly — a non-member falls through to `nil` → `cando_any?` returns false → 403. No legitimate own-tenant request regresses.
- `profiles`/`roles` now match the already-`tenant_authorize`d siblings `tenant/{users,groups}`.
- The OTP gate added to `tenant_authorize` mirrors the pre-existing `authorize` (json_api_controller.rb:380) verbatim; `otp_satisfied?` only denies MFA-enabled tokens that haven't provided OTP, so no non-MFA user is affected. It also correctly applies to the pre-existing `tenant_authorize` controllers (`access_control/{apps/tenant,tenant}/{users,groups}`) — the intended MFA-gap closure.
- `skip_before_action :authorize` is valid (registered at json_api_controller.rb:8), and the regression spec's static before-filter assertions hold.

No blocking issues. One non-blocking item below to confirm before merge.

app/controllers/api/v1/apps/tenants/organizations_controller.rb
neongrau: Confirmed against `identity-management-frontend` (`main`) — the org endpoints are app-admin-only, not a customer tenant-admin surface, so dropping `~/tenant.admin` does not regress any legitimate workflow:

- `src/pages/AdminUI/MenuDef.ts`: the path to org-unit management is **Apps** (`cando: "apps.reader"`) → an app → **Tenants** (`cando: "tenants.reader"`) → a tenant → *Organizational Units* (`/apps/:app/tenants/:tenant/ous` → `Apps/Tenants/OrgUnits`). Both navigation gates are app-admin candos.
- `OrganizationModel` (the `organizations` endpoints) is referenced only under `src/pages/AdminUI/...`; there is no customer-facing org-management UI outside the AdminUI.
- A plain `tenant.admin`-only user (e.g. the "Edit facility data" role) holds neither `apps.reader` nor `tenants.reader`, so cannot reach these screens. The endpoint was never a `tenant.admin`-only surface.

The same holds for the other Pattern B endpoints (`users/{roles,functionalities,actors}`, `organizations/{mappings,actor_roles}`, and the pickers) — they are all driven by the same AdminUI app-admin screens (`OrganizationRoleSelector`, `UserOrganizationSelector`, org/user forms under `AdminUI/Apps/Tenants`). Premise validated; good to merge from this angle.

SCjona (approved):

PR #271, opened 2026-08-03
Map users into standard_user when accepting a tenant invite

neongrau:
Fixes Samedis-care/samedis-care-issues#2403

## Problem

`Invite#_process_accept_tenant` had an inverted guard:

```ruby
standard_group.map_into!(user.actor) unless standard_group.is_a?(Actor)
```

The mapping was skipped exactly when the group **was** found — which is always, since
`standard_user` comes from `config/apps/samedis-care/actor_defaults/samedis-care.yml`
and gets `system: true` by default (`actor.rb:919`, the YAML sets no `system:` key).

Consequence: Samedis.care's "auto join facility" feature (`Staff#login_allowed`) linked
the staff record to the user account, but the user got no group membership and therefore
no `samedis-care-employee` role. `accept!` still set `accepted_at` / `done: true`, so the
next login did not retry it. Silent, no exception, no log entry.

Introduced by 635b04f during the `employee` → `standard_user` rename, which dropped the
preceding `raise 'NO EMPLOYEE GROUP FOUND' unless employee_group.is_a?(Actor)` and folded
the negated condition onto the `map_into!` line.

Reproduced in the test environment before the fix: `accept! => true`, `done = true`,
`Actors::Mapping` count `0`.

## Fix

1. `_process_accept_tenant` — guard flipped, `map_into!` runs when the group is present.
2. `accept!` — only marks an invite `done` when the processor reports success, so a failed
   attempt stays retryable instead of being accepted for nothing.

The group name and the `system: true` filter are unchanged; only the guard was wrong.

### Why a missing group reports instead of raising

The issue left this open, and the pre-635b04f code raised. Raising is not safe here:
`accept!` runs inside the login flow via `User#auto_accept_invites!` ←
`User#check_acceptances` (`user.rb:837`) and
`Api::V1::Devise::OmniauthCallbacksController` (`:53`), **with no rescue on either path**.
An exception would lock the user out of logging in entirely rather than out of one
facility — worse than the bug being fixed.

So a misconfigured tenant now gets `Sentry.capture_message(level: :error)` and the invite
is left unaccepted. Combined with change 2 this satisfies the issue's requirement that it
must not silently mark the invite accepted while granting nothing.

Both `_process_accept_*` methods (there are only two) end in an explicit `true`, so
change 2 alters nothing on the working paths — a spec covers the `access_control` path to
keep it that way.

## Tests

`spec/models/invite_spec.rb` — the first `Invite` spec in this repo. Real Mongoid
documents, following `spec/models/actors/tenant_spec.rb`. Builds tenant → organization →
`tenant_profiles` OU → group and covers:

- the user is mapped into `standard_user`, invite marked accepted
- no double mapping when accepted again
- missing group: not accepted, no raise, reported to Sentry
- `standard_user` that is not a system group is ignored
- `access_control` invites still get accepted

Verified to fail 6/8 on the unfixed code (including the core mapping example).

Suite: **106 examples / 19 failures** on this branch vs **98 / 19 on clean main** — the
same pre-existing failures in `social_auth_controller_spec` and
`omniauth_callbacks_controller_spec` (`ActionController::UrlGenerationError`, routing,
untouched here). All 8 new examples pass.

## Review notes / follow-ups (not in this PR)

- **Not asserted:** the resulting `samedis-care-employee` role. `Role belongs_to
  :actors_app` and `User#candos` go through a cached tenant-cando aggregation, so this
  needs the full samedis-care app config seeded into the test DB; it would test the role
  engine rather than this guard.
- **Not verified against production:** that every tenant actually has the `standard_user`
  system group. The Sentry path makes a wrong assumption visible instead of fatal.
- **Existing affected users are not repaired by this.** Their invites are `done: true`.
  `Staff#needs_auto_join_invite?` gates on `email_user.tenant_ids`, which never got
  populated, so a staff re-save creates a fresh invite and now works — but nothing
  re-saves on its own. A backfill belongs in `samedis-care-backend`.
- **Adjacent latent bug, deliberately out of scope:** `raise 'NO SUCH TENANT TO JOIN'` /
  `'NO SUCH USER'` (`invite.rb:111-114`) carry the same login-lockout hazard — a deleted
  tenant with a lingering `auto_accept` invite permanently 500s that user's login.

### DIFF
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -96,7 +96,10 @@ class Invite < ApplicationDocument
 
     action = "_process_accept_#{self.invitable_type}".to_sym
     raise "NO ACCEPT ACTION FOUND: #{action}" unless self.respond_to?(action)
-    self.send(action, *args)
+    # only burn the invite once processing actually granted something, otherwise
+    # it stays retryable on the next login instead of being accepted for nothing
+    return false unless self.send(action, *args)
+
     self.update_attributes(accepted_at: Time.now, done: true)
   end
 
@@ -117,7 +120,20 @@ class Invite < ApplicationDocument
                            .groups
                            .where(system: true, name: :standard_user).first
 
-    standard_group.map_into!(user.actor) unless standard_group.is_a?(Actor)
+    # A tenant without the system group `standard_user` is misconfigured - the group
+    # comes from config/apps/samedis-care/actor_defaults/samedis-care.yml. Do not raise
+    # here: this runs inside the login flow (User#check_acceptances) where an exception
+    # would lock the user out of logging in entirely. Report it and leave the invite
+    # unaccepted so the next login retries once the tenant is repaired.
+    unless standard_group.is_a?(Actor)
+      Sentry.capture_message(
+        "Invite#accept!: no system group 'standard_user' below tenant #{tenant_id} - user not joined",
+        level: :error
+      )
+      return false
+    end
+
+    standard_group.map_into!(user.actor)
 
     true
   end
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -168,6 +168,14 @@ class Invite < ApplicationDocument
   # OU at all. So a missing group is only a misconfiguration where the tenant's own app
   # defaults ask for it.
   def standard_user_expected?(tenant)
+    # Actors::Tenant#profiles_ou_defaults returns nil both when the app declares no
+    # tenant_profiles OU and when the tenant has no organization node at all
+    # (tenant.rb:100, and #organization filters by `available`, so a soft-deleted org tree
+    # gives nil on a live tenant). Only the first is an answer about the app; treating the
+    # second as "not expected" would burn the invite and report nothing, which is the
+    # silent no-op this whole change exists to remove.
+    return true if tenant.organization.nil?
+
     defaults = tenant.profiles_ou_defaults
     return false unless defaults.is_a?(Hash)
 
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -130,13 +130,20 @@ class Invite < ApplicationDocument
     # would lock the user out of logging in entirely. Report it and leave the invite
     # unaccepted so the next login retries once the tenant is repaired.
     unless standard_group.is_a?(Actor)
-      Sentry.capture_message(
-        "Invite#accept!: no system group 'standard_user' below tenant - user not joined",
-        level: :error,
-        tags: { tenant_id: tenant_id.to_s },
-        extra: { invite_id: id.to_s }
-      )
-      return false
+      if standard_user_expected?(tenant)
+        Sentry.capture_message(
+          "Invite#accept!: no system group 'standard_user' below tenant - user not joined",
+          level: :error,
+          tags: { tenant_id: tenant_id.to_s },
+          extra: { invite_id: id.to_s }
+        )
+        return false
+      end
+
+      # This app's tenants do not have that group, so there is nothing for this invite to
+      # grant and nothing to report. Count it as processed - leaving it unaccepted would
+      # re-run it on every login until it expires, for a condition that will never change.
+      return true
     end
 
     # same reasoning: `map_into!` raises on a missing or unpersisted actor (User#actor is
@@ -152,6 +159,22 @@ class Invite < ApplicationDocument
       return false
     end
 
+    true
+  end
+
+  # `standard_user` is a samedis-care actor default. invitable_type 'tenant' is settable for
+  # any app's tenant (both invitation controllers permit it), and other apps declare other
+  # groups - identity-management seeds `identity_management_admins` and no tenant_profiles
+  # OU at all. So a missing group is only a misconfiguration where the tenant's own app
+  # defaults ask for it.
+  def standard_user_expected?(tenant)
+    defaults = tenant.profiles_ou_defaults
+    return false unless defaults.is_a?(Hash)
+
+    Array(defaults['children']).any? { |child| child['name'].to_s.eql?('standard_user') }
+  rescue StandardError
+    # defaults unreadable: assume it was expected, so a real samedis-care misconfiguration
+    # still gets reported instead of being swallowed
     true
   end
 
--- a/app/models/invite.rb
+++ b/app/models/invite.rb
@@ -116,8 +116,12 @@ class Invite < ApplicationDocument
     user = get_user
     raise 'NO SUCH USER' unless user.is_a?(User)
 
+    # `available` matters: soft-deleting an ancestor cascades via a raw $set
+    # (`Actor#before_save`), which neither trips the system protection guard nor renames
+    # the group, so a group in a deleted subtree still matches name and system
     standard_group = tenant.descendants
                            .groups
+                           .available
                            .where(system: true, name: :standard_user).first
 
     # A tenant without the system group `standard_user` is misconfigured - the group
@@ -127,13 +131,26 @@ class Invite < ApplicationDocument
     # unaccepted so the next login retries once the tenant is repaired.
     unless standard_group.is_a?(Actor)
       Sentry.capture_message(
-        "Invite#accept!: no system group 'standard_user' below tenant #{tenant_id} - user not joined",
-        level: :error
+        "Invite#accept!: no system group 'standard_user' below tenant - user not joined",
+        level: :error,
+        tags: { tenant_id: tenant_id.to_s },
+        extra: { invite_id: id.to_s }
       )
       return false
     end
 
-    standard_group.map_into!(user.actor)
+    # same reasoning: `map_into!` raises on a missing or unpersisted actor (User#actor is
+    # optional) and can raise out of its own save!, and none of that may reach the login
+    begin
+      standard_group.map_into!(user.actor)
+    rescue StandardError => e
+      Sentry.capture_exception(
+        e,
+        tags: { tenant_id: tenant_id.to_s },
+        extra: { invite_id: id.to_s, group_id: standard_group.id.to_s }
+      )
+      return false
+    end
 
     true
   end
--- a/db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
+++ b/db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
@@ -12,7 +12,8 @@
 #   # 2. apply, after reviewing the report
 #   MANUAL=true APPLY=true RAILS_ENV=live bundle exec rails db:migrate
 #
-# Knobs: APPLY=true (write), LIMIT=n (cap repairs), TENANT_ID=<id> (single tenant).
+# Knobs: APPLY=true (write), LIMIT=n (cap repairs), TENANT_ID=<id> (single tenant),
+#        SINCE=YYYY-MM-DD or SINCE=all (default floor: 2025-04-16, the day the bug shipped).
 # Run once per cluster - the APAC cluster (RAILS_ENV=apac) has its own data.
 #
 # READ THIS BEFORE APPLYING
@@ -36,10 +37,23 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
 
   BUG_INTRODUCED = Time.utc(2025, 4, 16)
 
+  # Anything accepted before BUG_INTRODUCED has some other cause and is not repaired:
+  # the first live report found 68 of 94 predating the bug, going back to 2019, so this
+  # state is long-standing and not a fingerprint of the inverted guard. SINCE=all lifts
+  # the floor, SINCE=2024-01-01 moves it.
+  def self.since_floor
+    raw = ENV['SINCE'].to_s.strip
+    return nil if raw.casecmp('all').zero?
+    return BUG_INTRODUCED if raw.empty?
+
+    Time.parse(raw).utc
+  end
+
   def self.up
     apply = ENV['APPLY'].present?
     limit = ENV['LIMIT'].presence&.to_i
     only_tenant = ENV['TENANT_ID'].presence
+    since = since_floor
 
     scope = Invite.where(invitable_type: 'tenant', done: true)
     scope = scope.where(tenant_id: only_tenant) if only_tenant
@@ -97,6 +111,14 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
       stats[:affected] += 1
       stats[:affected_before_bug] += 1 if invite.accepted_at && invite.accepted_at < BUG_INTRODUCED
 
+      # counted as affected for the report, but out of repair scope
+      if since && (invite.accepted_at.nil? || invite.accepted_at < since)
+        stats[:skip_before_since] += 1
+        next
+      end
+
+      stats[:in_repair_scope] += 1
+
       if limit && repaired.size >= limit
         stats[:over_limit] += 1
         next
@@ -118,7 +140,7 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
       end
     end
 
-    report(stats, months, apply)
+    report(stats, months, apply, since)
 
     # keep the migration unrecorded so the real run still has something to do
     unless apply
@@ -147,7 +169,7 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
     Actors::Tenant.available.where(_id: oid).first
   end
 
-  def self.report(stats, months, apply)
+  def self.report(stats, months, apply, since = nil)
     say '=' * 76
     say "accepted tenant invites with no standard_user mapping, by acceptance month"
     say "(the inverted guard shipped with 635b04f on #{BUG_INTRODUCED.strftime('%Y-%m-%d')})"
@@ -155,11 +177,20 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
       say format('  %-8s %5d %s', month, months[month], '#' * [months[month], 50].min)
     end
     say '-' * 76
-    %i[scanned affected affected_before_bug repaired failed already_mapped
-       skip_has_other_access skip_tenant_without_group skip_tenant_gone
-       skip_user_gone skip_user_without_actor over_limit].each do |key|
+    %i[scanned affected affected_before_bug skip_before_since in_repair_scope
+       repaired failed already_mapped skip_has_other_access
+       skip_tenant_without_group skip_tenant_gone skip_user_gone
+       skip_user_without_actor over_limit].each do |key|
       say format('  %-26s %6d', key, stats[key])
     end
+    say '-' * 76
+    say(if since
+          "repair scope: accepted on/after #{since.strftime('%Y-%m-%d')} " \
+            "(#{stats[:in_repair_scope]} of #{stats[:affected]} affected). " \
+            'SINCE=all lifts the floor, SINCE=YYYY-MM-DD moves it.'
+        else
+          "repair scope: NO date floor, all #{stats[:affected]} affected rows in scope"
+        end)
     say '=' * 76
     say(apply ? "repaired #{stats[:repaired]} mapping(s)" : 'DRY RUN - nothing written')
   end
--- a/db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
+++ b/db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
@@ -13,7 +13,8 @@
 #   MANUAL=true APPLY=true RAILS_ENV=live bundle exec rails db:migrate
 #
 # Knobs: APPLY=true (write), LIMIT=n (cap repairs), TENANT_ID=<id> (single tenant),
-#        SINCE=YYYY-MM-DD or SINCE=all (default floor: 2025-04-16, the day the bug shipped).
+#        SINCE=YYYY-MM-DD or SINCE=all (default floor: 2025-04-16, the day the bug shipped),
+#        EMAILS=a@x.de,b@y.de (repair only these, after confirming them on the samedis side).
 # Run once per cluster - the APAC cluster (RAILS_ENV=apac) has its own data.
 #
 # READ THIS BEFORE APPLYING
@@ -54,6 +55,10 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
     limit = ENV['LIMIT'].presence&.to_i
     only_tenant = ENV['TENANT_ID'].presence
     since = since_floor
+    # EMAILS=a@x.de,b@y.de restricts the repair to an explicitly confirmed set. Rows
+    # outside it are still reported, so the full picture stays visible.
+    only_emails = ENV['EMAILS'].to_s.split(/[,\s]+/).map { |e| e.downcase.strip }
+                               .reject(&:empty?).presence
 
     scope = Invite.where(invitable_type: 'tenant', done: true)
     scope = scope.where(tenant_id: only_tenant) if only_tenant
@@ -118,9 +123,15 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
         next
       end
 
+      if only_emails && !only_emails.include?(user.email.to_s.downcase)
+        stats[:skip_not_in_emails] += 1
+        next
+      end
+
       stats[:in_repair_scope] += 1
       in_scope << { email: user.email, tenant: tenant.name.to_s,
-                    accepted_at: invite.accepted_at, invite_id: invite.id.to_s }
+                    tenant_id: tenant.id.to_s, accepted_at: invite.accepted_at,
+                    invite_id: invite.id.to_s }
 
       if limit && repaired.size >= limit
         stats[:over_limit] += 1
@@ -180,8 +191,8 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
       say format('  %-8s %5d %s', month, months[month], '#' * [months[month], 50].min)
     end
     say '-' * 76
-    %i[scanned affected affected_before_bug skip_before_since in_repair_scope
-       repaired failed already_mapped skip_has_other_access
+    %i[scanned affected affected_before_bug skip_before_since skip_not_in_emails
+       in_repair_scope repaired failed already_mapped skip_has_other_access
        skip_tenant_without_group skip_tenant_gone skip_user_gone
        skip_user_without_actor over_limit].each do |key|
       say format('  %-26s %6d', key, stats[key])
@@ -207,6 +218,11 @@ class RepairMissingStandardUserMappings < Mongoid::Migration
       say '-' * 76
       mails = in_scope.map { |r| r[:email].presence }
       say "emails only: #{mails.compact.sort.join(' ')}"
+      # email alone is ambiguous for a customer running many tenants - the same person can
+      # be staff at several, and only one of them may be the one that lost access
+      say 'email|tenant pairs: ' +
+          in_scope.select { |r| r[:email].present? }
+                  .map { |r| "#{r[:email]}|#{r[:tenant_id]}" }.sort.join(' ')
       # without an address these cannot be matched against the samedis-care staff list,
       # so they need deciding by hand rather than slipping through with the rest
       blank = in_scope.reject { |r| r[:email].present? }
--- a/db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
+++ b/db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
@@ -4,7 +4,7 @@
 # so `Staff#login_allowed` linked the account while the user got no access and no
 # samedis-care-employee role. See Samedis-care/samedis-care-issues#2403
 #
-# Lives in db/migrate/manual, so it only runs when MANUAL is set:
+# Lives in db/migrate_manual, so it only runs when MANUAL is set:
 #
 #   # 1. report only, writes nothing, is NOT recorded as migrated (default)
 #   MANUAL=true RAILS_ENV=live bundle exec rails db:migrate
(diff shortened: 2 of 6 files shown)


app/models/invite.rb
```diff
@@ -117,7 +120,20 @@ def _process_accept_tenant(*_)
                            .groups
                            .where(system: true, name: :standard_user).first
```
SCjona, line 125: The lookup does not exclude soft-deleted actors, and now that the mapping actually runs this matters.

Every other group lookup in the codebase filters with `available`: `user.rb:946` (`app.organization.descendants.groups.available`), `actors/tenant.rb:216`, `apps/groups_controller.rb:5`, `apps/tenants/picker/users/groups_controller.rb:5`, etc.

It is reachable, not theoretical: soft-deleting an ancestor cascades via `record.descendants.set(deleted: record.deleted?)` (`actor.rb:355`), which is a raw `$set` — it bypasses both the `system` protection guard (`actor.rb:331`) and `name_uniquely_disabled!` (`actor.rb:353`, which is applied only to the record itself, not to the cascaded descendants). So a group inside a soft-deleted tenant/org subtree keeps `name: "standard_user"` and `system: true` and still matches this criteria. Combined with the unordered `.first`, a tenant whose org tree was deleted and rebuilt can have both a deleted and a live `standard_user` group, and this picks whichever Mongo returns first.

Result in that case: `map_into!` succeeds against a deleted group, `accept!` marks the invite `done: true`, and the user ends up with no effective role — the exact silent no-op this PR fixes.

```suggestion
                           .available
                           .where(system: true, name: :standard_user).first
```

app/models/invite.rb
```diff
@@ -117,7 +120,20 @@ def _process_accept_tenant(*_)
...
 
-    standard_group.map_into!(user.actor) unless standard_group.is_a?(Actor)
+    # A tenant without the system group `standard_user` is misconfigured - the group
+    # comes from config/apps/samedis-care/actor_defaults/samedis-care.yml. Do not raise
+    # here: this runs inside the login flow (User#check_acceptances) where an exception
+    # would lock the user out of logging in entirely. Report it and leave the invite
+    # unaccepted so the next login retries once the tenant is repaired.
+    unless standard_group.is_a?(Actor)
+      Sentry.capture_message(
+        "Invite#accept!: no system group 'standard_user' below tenant #{tenant_id} - user not joined",
```
SCjona: Interpolating `tenant_id` into the message body gives every tenant its own Sentry fingerprint, so this fragments into one issue per affected tenant instead of one issue with N events.

That interacts with change 2: because the invite is no longer burned, `auto_accept_invites!` re-runs this on **every login** of every affected user until `valid_until` (up to 30 days). Keeping the identifier out of the message and in tags/extra keeps it one groupable issue with a per-tenant breakdown, and lets Sentry deduplicate:

```ruby
Sentry.capture_message(
  "Invite#accept!: no system group 'standard_user' below tenant - user not joined",
  level: :error,
  tags: { tenant_id: tenant_id.to_s },
  extra: { invite_id: id.to_s, email: email }
)
```

The spec matcher (`with(/standard_user/, hash_including(level: :error))`) still passes with this shape.

app/models/invite.rb
```diff
@@ -117,7 +120,20 @@ def _process_accept_tenant(*_)
...
+        level: :error
+      )
+      return false
+    end
+
+    standard_group.map_into!(user.actor)
```
SCjona: Minor, and consistent with the reasoning you already applied above: this call is itself a newly-live raise site inside the login flow.

`belongs_to :actor` on User is `optional: true` (`user.rb:77`) and `map_into!` raises `Can only map actors!` on nil / `Can only map persisted actors!` on a new record (`actor.rb:772-773`); `mapped_actor.save!` (`actor.rb:788`) can also raise on an Actor validation. Before this PR none of that was reachable, because `map_into!` never executed. The codebase already treats a missing actor as possible — `User#apps` does `raise "MISSING ACTOR FOR: ..."` (`user.rb:603`).

Since the point of the missing-group branch is that nothing here may raise into `check_acceptances`, consider giving the mapping the same treatment:

```ruby
begin
  standard_group.map_into!(user.actor)
rescue StandardError => e
  Sentry.capture_exception(e, tags: { tenant_id: tenant_id.to_s })
  return false
end
```

Your call — this restores long-standing pre-`635b04f` behaviour, so leaving it as-is is defensible.

SCjona (commented):
The core fix is correct and I verified the claim in the PR description against the introducing commit: `635b04f` dropped `raise 'NO EMPLOYEE GROUP FOUND' unless employee_group.is_a?(Actor)` and folded the negation onto the `map_into!` line, so the mapping was skipped precisely when the group existed. This PR restores the pre-`635b04f` semantics minus the raise, and the reasoning for reporting instead of raising is sound — `accept!` runs from `User#check_acceptances` (`user.rb:837`) and `Api::V1::Devise::OmniauthCallbacksController:53` with no rescue on either path, so a raise there is a login lockout, not a facility lockout.

The `accept!` change is safe: both `_process_accept_*` methods end in an explicit `true`, neither caller (`user.rb:822`, `Api::V1::User::InvitationsController:23`) inspects the return value, and the early `return false` in `_process_accept_tenant` happens before any mutation, so there is no half-applied state.

Three non-blocking suggestions inline, plus one note that has no diff line to attach to:

**`Api::V1::User::InvitationsController#update` (`invitations_controller.rb:20-28`)** ignores the return value and renders 200 regardless. On the manual-accept path a misconfigured tenant now produces a 200 with an invite that silently stayed pending, where before it produced a 200 with an invite that was (wrongly) marked done. Not a regression, but `accept!` now has a meaningful boolean — worth turning a `false` into a 422 in a follow-up so the frontend can tell the difference.

Agreed on the out-of-scope items in the description; the `NO SUCH TENANT TO JOIN` / `NO SUCH USER` raises carry the same lockout hazard and are correctly left for a separate change.

app/models/invite.rb
neongrau: Verified and applied in 6b366a4 — this was a real defect in my change, not a hypothetical.

Confirmed every link: `available` is `criteria.not.where(deleted: true)` (`application_document.rb:136`), the cascade is a raw `$set` (`actor.rb:355`), the `system`+`deleted` abort is a `before_save` guard (`actor.rb:331`) so the `$set` bypasses it, and `name_uniquely_disabled!` (`actor.rb:353`) applies only to the record itself.

The `.first` ordering concern reproduces empirically. The new spec `still maps into a live group next to the deleted one` fails without the filter: Mongo returned the **deleted** group ahead of the live one, so `map_into!` targeted the deleted group and the live one got no mapping — exactly the silent no-op this branch exists to fix.

app/models/invite.rb
neongrau: Applied in 6b366a4, with one deliberate deviation: I left `email` out of `extra` and kept only `invite_id`, since the invite id is enough to reach the record and the email is personal data going to a third-party service. `tags: { tenant_id: }` and `extra: { invite_id: }` are in.

Your point about the repeat rate is the sharper half — with the invite no longer burned this fires on every login of every affected user until `valid_until` (up to 30 days), so grouping matters more than it would have before change 2.

app/models/invite.rb
neongrau: Adopted in 6b366a4 rather than left as-is — you are right that it is the same hazard, and "restores pre-`635b04f` behaviour" is not a good enough reason when that behaviour was itself only reachable through the bug.

Verified reachable: with the rescue removed, the new spec `reports and stays retryable when the user has no actor to map` fails with `Can only map actors!` from `actor.rb:772`, raising straight through `accept!` into the login flow.

One note on the shape: a partial `map_into!` (mapping saved, a later step raising) leaves the invite retryable while the mapping exists — the next login hits `return mapped_actor if mapped_actor.persisted?` (`actor.rb:786`), gets a truthy result, and the invite is accepted. So it self-heals rather than sticking.

app/models/invite.rb
```diff
@@ -113,11 +116,41 @@ def _process_accept_tenant(*_)
...
 
-    standard_group.map_into!(user.actor) unless standard_group.is_a?(Actor)
+    # A tenant without the system group `standard_user` is misconfigured - the group
+    # comes from config/apps/samedis-care/actor_defaults/samedis-care.yml. Do not raise
+    # here: this runs inside the login flow (User#check_acceptances) where an exception
+    # would lock the user out of logging in entirely. Report it and leave the invite
+    # unaccepted so the next login retries once the tenant is repaired.
+    unless standard_group.is_a?(Actor)
+      Sentry.capture_message(
+        "Invite#accept!: no system group 'standard_user' below tenant - user not joined",
```
SCjona: `standard_user` is a samedis-care actor-default, but this branch is app-agnostic while `invitable_type: 'tenant'` is client-settable for any app's tenant: `Api::V1::App::Tenant::InvitationsController#params_create` permits `:invitable_type` and forces `app: current_app` (`app/tenant/invitations_controller.rb:26,43`), and `Api::V1::User::InvitationsController#params_create` permits it too (`user/invitations_controller.rb:55`).

For a tenant under a non-samedis app — `identity-management` itself seeds `identity_management_admins`, not `standard_user` (`config/apps/identity-management/actor_defaults/identity-management.yml`) — the group is legitimately absent, so this reports `level: :error` claiming a misconfiguration that isn't one. Combined with change 2 the invite is no longer burned, so `auto_accept_invites!` re-runs it and re-reports on **every login** of that user until `valid_until` (up to 30 days), after which it expires silently.

Functionally this is no worse than before (no mapping either way), so it is alert quality rather than a regression. Worth considering a guard so the error only fires where `standard_user` is actually expected — e.g. checking `app` before reporting — or wording that doesn't assert misconfiguration. Your call on whether cross-app tenant invites exist in practice; the two permit lists say they are reachable.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
```diff
@@ -0,0 +1,239 @@
...
+  # Anything accepted before BUG_INTRODUCED has some other cause and is not repaired:
+  # the first live report found 68 of 94 predating the bug, going back to 2019, so this
+  # state is long-standing and not a fingerprint of the inverted guard. SINCE=all lifts
+  # the floor, SINCE=2024-01-01 moves it.
+  def self.since_floor
+    raw = ENV['SINCE'].to_s.strip
+    return nil if raw.casecmp('all').zero?
+    return BUG_INTRODUCED if raw.empty?
+
+    Time.parse(raw).utc
```
SCjona: `Time.parse` is lenient in a way that matters for a knob typed by hand on a live console. `SINCE=nonsense` aborts the whole `db:migrate` run with a bare `ArgumentError` and no hint that `SINCE` was the cause, and an input that isn't a full date can be parsed as a time-of-day rather than a year (`Time.parse('2025')`), silently yielding a floor of today — which makes the run report `in_repair_scope 0` and read as "nothing to do" rather than "your SINCE was wrong".

The report does echo the resolved floor (`repair scope: accepted on/after …`), so it is detectable, but a format check would make it self-explanatory:

```ruby
raise "SINCE must be YYYY-MM-DD or 'all', got #{raw.inspect}" unless raw.match?(/\A\d{4}-\d{2}-\d{2}\z/)

Time.parse(raw).utc
```

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
```diff
@@ -0,0 +1,239 @@
+# Repairs users who accepted a `tenant` invite while Invite#_process_accept_tenant had
+# the inverted guard (introduced 2025-04-16 by 635b04f, fixed in #271). Those invites were
+# marked `done: true` without ever mapping the user into the tenant's standard_user group,
+# so `Staff#login_allowed` linked the account while the user got no access and no
+# samedis-care-employee role. See Samedis-care/samedis-care-issues#2403
+#
+# Lives in db/migrate_manual, so it only runs when MANUAL is set:
+#
+#   # 1. report only, writes nothing, is NOT recorded as migrated (default)
+#   MANUAL=true RAILS_ENV=live bundle exec rails db:migrate
```
SCjona: "report only, writes nothing" is true of this migration but not of this command. `MANUAL` appends to the path rather than replacing it (`config/application.rb:95,105`), so with `MANUAL=true` the path is `['db/migrate', 'config/apps/*/migrate', 'db/migrate_manual']` and `rails db:migrate` applies any *pending regular* migrations first, then raises here.

On a cluster that is already fully migrated that is a no-op, which is presumably the expected state for a live run — but one line in this block would tell whoever runs it on a cluster that is behind (or on APAC, which has never had this migration) to check `db:migrate:status` first. Same reason worth noting: the dry-run raise cancels anything ordered after `20260803081500`.

Minor related edge: with both `INDEXES` and `MANUAL` set the path becomes `['db/migrate/indexes', 'db/migrate_manual']`, because the `INDEXES` branch assigns and the `MANUAL` branch appends.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
```diff
@@ -0,0 +1,239 @@
...
+        stats[:skip_not_in_emails] += 1
+        next
+      end
+
+      stats[:in_repair_scope] += 1
+      in_scope << { email: user.email, tenant: tenant.name.to_s,
+                    tenant_id: tenant.id.to_s, accepted_at: invite.accepted_at,
+                    invite_id: invite.id.to_s }
+
+      if limit && repaired.size >= limit
```
SCjona, line 173: `in_scope` and `stats[:in_repair_scope]` are both incremented above this check, so in a `LIMIT`-capped `APPLY=true` run the report's `rows in repair scope (N) - cross-check these …` block, the `emails only:` line and the `email|tenant pairs:` line all list rows that were **not** repaired. Since that output is the audit trail for a change that grants access, it would be better if the list matched what was actually written — `over_limit` is printed, so the discrepancy is derivable, but only if someone notices the two numbers disagree.

Cheapest fix is to move the `in_scope <<` push below this limit check (leaving `stats[:in_repair_scope]` as the "passes the filters" counter), or to mark the over-limit rows in the printed list.

Secondary, same loop: the group lookup at line 83 plus the two `Actors::Mapping` queries run once per invite, and accepted tenant invites cluster heavily by tenant. Memoizing the group by `invite.tenant_id` in a hash would cut most of the query volume on the full-cluster scan — the first report showed 94 affected, but `scanned` is presumably a much larger number.

SCjona (commented):
Re-reviewed at `4565b5c`. All three findings from the previous round are correctly applied, and I re-verified each against the code rather than the reply:

- `available` is in place at `invite.rb:124`; `ApplicationDocument.available` is `criteria.not.where(deleted: true)` (`application_document.rb:136`) and the cascade at `actor.rb:355` is a raw `$set`, so the filter is load-bearing exactly as described.
- Sentry grouping: identifier out of the message body, `tags: { tenant_id: }` / `extra: { invite_id: }`. Leaving `email` out is the right call, and `capture_message`/`capture_exception` both accept these keys in sentry-ruby 6.6.
- The `rescue StandardError` around `map_into!` is there. I confirmed the raise sites: `map_into!` raises on a nil/new-record actor (`actor.rb:772-773`) and `belongs_to :actor` is `optional: true` (`user.rb:77`). I also chased the follow-on worry that a repaired mapping might not become visible to the user — it does: `Actors::Mapping` has `after_save :user_cache_expire!` (`actors/mapping.rb:7`) which calls `map_actor.user.cache_expire!`, so `tenant_candos_cached` is invalidated by the mapping save itself, independently of the `cache_expire!` in `User#auto_accept_invites!`. That matters because `Actor#cache_expire!` on the user's own actor is a no-op here (`actor.rb:626` looks for mappings *below* that actor), and `User#candos` otherwise only refreshes at the `beginning_of_day` boundary.

The `db/migrate_manual` relocation is right, and the reasoning in the comment is verifiable in this repo: `db/migrate/indexes/20231129144329_*` is already picked up by the plain `db/migrate` path for exactly the same globbing reason, which is why a `db/migrate/manual` subdirectory would have collided.

The migration itself is careful — the irreversibility note, the `skip_has_other_access` guard, the `2025-04` floor with the report showing what the floor excluded, and the `email|tenant` pairs for the cross-check all address the "cannot distinguish bug from deliberate revocation" problem honestly. I checked two things that could have made the scan unsafe and both are fine: `Invite` has no `deleted` field, so there is no soft-deleted/declined invite that could be swept in (`JsonApiController#destroy` hard-destroys it), and a soft-deleted mapping under a live `available` group is not reachable, because the delete/restore cascade at `actor.rb:355-356` is symmetric and `before_save` hard-deletes any mapping flagged deleted (`actor.rb:341`) — so the `already_mapped` check not filtering `available` is not a gap.

Four non-blocking notes below, none of which need to hold the merge.

app/models/invite.rb
neongrau: Applied in 9ef2f90, and you were right that this is more than alert quality — I had turned a silent no-op into a permanent per-login retry for those tenants.

Verified both halves of the claim: `invitable_type` is permitted in `user/invitations_controller.rb:55` and `app/tenant/invitations_controller.rb:43`, and `config/apps/identity-management/actor_defaults/identity-management.yml` contains zero occurrences of `standard_user` — it seeds `identity_management_admins` and declares no `tenant_profiles` OU at all.

Rather than keying on the app name, the check asks the tenant’s own defaults whether the group is declared (`Actors::Tenant#profiles_ou_defaults`). Where it is not, there is nothing to grant, so the invite is accepted as it was before this PR — which also restores pre-`635b04f` behaviour for those apps instead of inventing a new retry loop for them. Unreadable defaults fall back to "expected", so a genuine samedis-care misconfiguration still reports rather than being swallowed.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
neongrau: Applied in 9ef2f90 — and the spec I added for it caught a second, quieter bug in the same method.

The format check is in. But `Time.parse("2024-01-01").utc` anchors the date in the server’s local zone, so on CET it resolved to `2023-12-31 23:00 UTC` — an hour of drift against `BUG_INTRODUCED`, which is built with `Time.utc`. An explicit `SINCE` therefore meant something slightly different from the default floor. Now built as `Time.utc(*raw.split("-").map(&:to_i))`, which the regex already guarantees is safe.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
neongrau: Applied in 9ef2f90, and this one had already reached the operator: the command I put in the header — and passed on directly — was `db:migrate`.

Switched the documented invocation to `db:migrate:up VERSION=20260803081500`, which runs only this migration. Your APAC point is the sharp end of it: that cluster has never seen this file, so "report only, writes nothing" would have applied every pending regular migration there first. The `INDEXES`+`MANUAL` interaction is noted in the header too.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
neongrau: Both applied in 9ef2f90.

`in_scope <<` moved below the limit check, so the printed list, the `emails only:` line and the `email|tenant pairs:` line describe what was written rather than what passed the filters. `stats[:in_repair_scope]` stays the "passed the filters" counter, and a spec asserts the split (`in_repair_scope 1`, `over_limit 1`, empty list) under `LIMIT=0`.

Group lookup memoized per tenant id. Your guess about the ratio was right — the live run scanned 872 invites, and the affected rows turned out to sit in a single customer group’s tenants, so almost all of those lookups were repeats.

app/models/invite.rb
```diff
@@ -113,12 +116,65 @@ def _process_accept_tenant(*_)
...
+  end
 
+  # `standard_user` is a samedis-care actor default. invitable_type 'tenant' is settable for
+  # any app's tenant (both invitation controllers permit it), and other apps declare other
+  # groups - identity-management seeds `identity_management_admins` and no tenant_profiles
+  # OU at all. So a missing group is only a misconfiguration where the tenant's own app
+  # defaults ask for it.
+  def standard_user_expected?(tenant)
+    defaults = tenant.profiles_ou_defaults
+    return false unless defaults.is_a?(Hash)
```
SCjona, line 180: `profiles_ou_defaults` returns `nil` for two different reasons, and this line collapses them into "not expected":

```ruby
def profiles_ou_defaults
  @profiles_ou_defaults ||= begin
    return nil unless organization                      # (a) no organization
    organization.defaults[:children].detect do
      it['name'].eql?('tenant_profiles')
    end.try(:with_indifferent_access)                   # (b) app declares no tenant_profiles
  end
end
```

(b) is the case this guard is for — `identity-management` — and handling it as "not expected" is right. (a) is not an answer about the app at all, and taking the `return false` path means `_process_accept_tenant` returns `true`, the invite is marked `done: true`, and nothing is granted and nothing is reported. That is precisely the silent no-op class this PR exists to remove, arriving through the guard added to fix the noise.

Reachable because `Actors::Tenant#organization` is `children.available.where(_type: Actors::Organization).first` (`tenant.rb:62`) — an `available` filter, same as the one you added at `invite.rb:124`. A tenant whose organization node was soft-deleted (the "org tree deleted and rebuilt" mid-state from the previous round's finding) therefore has `organization == nil` while the tenant itself is live. `Actor.tenants.find(tenant_id)` at `:113` has no `available` filter either, so a soft-deleted tenant lands here too.

The `rescue StandardError` fallback does not cover it: the `organization`-blank path returns a clean `nil` rather than raising. Contrast with the app-less case, which does hit the rescue — `organization.defaults` is `{}` when `app.blank?` (`organization.rb:19`), so `{}[:children].detect` raises `NoMethodError` and correctly resolves to "expected". That asymmetry is also what makes the spec `when the tenant has no standard_user group` pass: it reaches the Sentry branch through the rescue, not through this line — so the current specs would not catch a regression here either.

Cheapest fix is to ask the distinguishing question directly:

```ruby
  def standard_user_expected?(tenant)
    # no organization at all is not an answer about the app - fall through to the same
    # fail-safe as unreadable defaults, so the tenant gets reported rather than burned
    return true if tenant.organization.nil?

    defaults = tenant.profiles_ou_defaults
    return false unless defaults.is_a?(Hash)
```

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
```diff
@@ -0,0 +1,266 @@
...
+# then cancel anything ordered after this one. On a cluster that is behind, or on APAC which
+# has never seen this file, that is not a report but a deploy. Check `db:migrate:status`
+# first if unsure.
+#
+#   # 1. report only, writes nothing, is NOT recorded as migrated (default)
+#   MANUAL=true VERSION=20260803081500 RAILS_ENV=live bundle exec rails db:migrate:up
+#
+#   # 2. apply, after reviewing the report and confirming the set on the samedis-care side
+#   MANUAL=true APPLY=true EMAILS=a@x.de,b@y.de VERSION=20260803081500 \
+#     RAILS_ENV=live bundle exec rails db:migrate:up
```
SCjona, line 21: The documented rollout is a batched one — `EMAILS=` restricts the write to a confirmed set, and the live status you posted has 10 of 26 in-scope rows confirmed so far. But this command records the migration, and `db:migrate:up` is a no-op once a version is recorded:

```ruby
# Mongoid::Migrator#run
target = migrations.detect { |m| m.version == @target_version }
raise UnknownMigrationVersionError.new(@target_version) if target.nil?
unless (up? && migrated.include?(target.version.to_i)) || (down? && !migrated.include?(target.version.to_i))
  target.migrate(@direction)
  record_version_state_after_migrating(target.version)
end
```

So the second batch — the remaining 16 once the samedis-care side confirms them — silently does nothing. No `say` output, no report, no error; the operator sees the task exit cleanly and has no signal distinguishing "nothing left to repair" from "never ran". `self.down` correctly raises `IrreversibleMigration`, so `db:migrate:down` is not a way back either.

The dry run is unaffected: it raises before `record_version_state_after_migrating`, which is exactly why step 1 stays repeatable.

Worth a line in this header covering how to run the later batches, e.g.:

```
#   # 3. later batches: the version is recorded by the first APPLY run and db:migrate:up
#   #    then silently does nothing. Run subsequent batches directly instead:
#   MANUAL=true APPLY=true EMAILS=c@x.de RAILS_ENV=live bundle exec rails runner \
#     'load Rails.root.join("db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb").to_s
#      RepairMissingStandardUserMappings.up'
```

(`MANUAL` is not needed on that path, since nothing consults the migration path.) Alternatively, apply the full confirmed set in one run and treat the recorded version as final. Either is fine — what matters is that the current header reads as if step 2 were repeatable.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
```diff
@@ -0,0 +1,266 @@
...
+
+      tenant = tenant_for(invite)
+      if tenant.nil?
+        stats[:skip_tenant_gone] += 1
+        next
+      end
+
+      # accepted tenant invites cluster heavily by tenant, so memoize rather than
+      # repeating this lookup for every invite of the same tenant
+      groups_by_tenant[tenant.id] ||= tenant.descendants.groups.available
```
SCjona: `||=` re-runs the query whenever the value is `nil`, which here is exactly the `skip_tenant_without_group` case — a tenant that is missing the group pays the `descendants.groups.available` lookup once per accepted invite, i.e. the full un-memoized cost for the population most likely to have many invites clustered on one tenant.

`fetch` with a block keyed on presence avoids it:

```ruby
      group = groups_by_tenant.fetch(tenant.id) do
        groups_by_tenant[tenant.id] = tenant.descendants.groups.available
                                            .where(system: true, name: :standard_user).first
      end
```

Minor — the scan is a one-off and 872 invites is small — but it is the one case the memoization was meant to cover and currently misses.

SCjona (commented):
Re-reviewed at `9ef2f90`. All four notes from the second round are applied, and I re-verified each against the code rather than the replies:

- **`standard_user_expected?`** — the claim holds. `Actors::Organization#defaults` builds from the live app template tree (`Actor#get_app_orga`, `actor.rb:215`), `treeify` keeps `children` on the top-level nodes (the depth-1 OUs), so `profiles_ou_defaults['children']` really is the group list and `child['name']` resolves through `with_indifferent_access` (`tenant.rb:104`). For `identity-management` there is no `tenant_profiles` node, `detect` returns nil, and the branch accepts the invite as it did before this PR.
- **`SINCE` / `Time.utc`** — confirmed: `Time.parse('2024-01-01').utc` is zone-anchored and would drift against `BUG_INTRODUCED`, which is `Time.utc`. `Time.utc(*raw.split('-').map(&:to_i))` after the `\A\d{4}-\d{2}-\d{2}\z` check is correct, and an out-of-range date still raises loudly rather than silently.
- **`db/migrate_manual` + `db:migrate:up`** — the relocation and the documented invocation are both right. I re-checked the globbing rationale: `db/migrate/indexes/20231129144329_*` is already reachable from the plain `db/migrate` path for the same reason, so a `db/migrate/manual` subdirectory would indeed have collided once appended.
- **`in_scope` below the limit check** — applied; the printed list, `emails only:` and `email|tenant pairs:` now describe writes rather than filter passes.

Three findings below. The first is a behaviour gap in the new guard, the second is an operational trap in the documented rollout for the repair migration — worth resolving before the `APPLY=true` runs, but neither needs to block the merge of the guard fix itself.

One smaller thing with no good line to attach to: in the `APPLY=true` loop a row whose `map_into!` raises is already in `in_scope`, so it appears in the "rows in repair scope … cross-check these" list even though nothing was written for it. `stats[:failed]` makes it derivable, same as `over_limit` did before; moving the push after the `begin`/`rescue` (or annotating failed rows) would make the list literally the write log it is described as.

app/models/invite.rb
neongrau: Fixed in a1e5bff. This is the sharpest finding of the three rounds — the guard I added to remove the noise reintroduced the exact silent no-op the PR exists to remove, for a narrower population.

Verified every step of the reasoning: `organization.defaults` is `{}` when `app.blank?` (`organization.rb:19`), so `{}[:children].detect` raises `NoMethodError` and the app-less case resolves to "expected" through the rescue — while `return nil unless organization` (`tenant.rb:100`) returns cleanly and did not. `#organization` filters by `available` (`tenant.rb:62`), so the soft-deleted-org-tree mid-state from round one produces exactly this on a live tenant.

Your point about the spec is the part I would have missed: the existing `when the tenant has no standard_user group` example passes through the rescue, not through this distinction, so it gave false confidence. Added a spec that stubs `#organization` to nil, and confirmed it fails with `got true` when the guard line is removed.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
neongrau: Documented in a1e5bff — and this one is not hypothetical, it cost us an hour on live earlier today.

After the first `APPLY` run recorded the version, `MANUAL=true VERSION=20260803081500 rails db:migrate:up` printed the migrations_path banner and nothing else. Exactly what you describe: no `say` output, no report, no error, indistinguishable from a clean run with nothing to do. I initially misread it as my dry-run raise having failed to prevent the recording, until the `DataMigration` ObjectId dated the record to the minute of the `APPLY` run.

The header now covers the batched case with the runner invocation, which ignores the recorded version and is also the repeatable way to re-report afterwards — that is how the post-apply state was measured. Confirmed `MANUAL` is unnecessary there.

db/migrate_manual/20260803081500_repair_missing_standard_user_mappings.rb
neongrau: Both applied in a1e5bff.

`fetch` with a block instead of `||=` — right that it missed precisely the population it was meant to cover, since `nil` is the `skip_tenant_without_group` value.

And the note without a line to attach to: `in_scope` is now pushed inside the success path of the `begin`/`rescue` and in the dry-run branch, so a row whose `map_into!` raised no longer appears in a list the header describes as the write log. `stats[:failed]` still counts it, and the `FAILED invite=… tenant=… user=…` line still identifies it.

SCjona (approved):

PR #274, opened 2026-08-11
Block untrusted libvips loaders and validate image upload MIME types

neongrau:
Fixes Samedis-care/samedis-care-issues#2451 — the IMB half of the libvips problem behind Samedis-care/samedis-care-issues#2412. Same fix as samedis-care-backend#4108.

## What this is about

CVE-2026-66066 / GHSA-xr9x-r78c-5hrm names `activestorage`. We don't load Active Storage (`config/application.rb:8`), so that path isn't exploitable — but the vulnerable component is **libvips**, and we call it for user/actor image derivatives and for the mail logo. libvips flags some loaders/savers as "unfuzzed": feeding them attacker-supplied files can disclose arbitrary files readable by the process, potentially RCE.

## Changes

1. `config/initializers/vips.rb` — `Vips.block_untrusted(true)`, plus a boot-time guard that refuses libvips < 8.13 (older versions can't block; our `ruby:4.0.1-trixie` image is fine).
2. `ImageUploader` and `ActorImageUploader` — `validation_helpers` plus a `png/jpeg/svg` MIME allowlist. Both had **no validations at all**, so arbitrary uploaded bytes reached libvips with the loader chosen by content sniffing.
3. Specs for both.

## Two things worth knowing

**image_processing already blocks untrusted loaders on require.** `image_processing/vips` calls `block_untrusted` itself, and both uploaders require it — so this is partly mitigated today. What it does not cover is `DeviseMailer#logo` (`app/mailers/devise_mailer.rb:209`), which hands the app's configured `mailer.logo_b64` straight to `Vips::Image.new_from_file`. The initializer removes the load-order dependency.

Side effect of that existing gem behavior: **SVG uploads have been failing since IMB picked up image_processing 2.x** (`Vips::Error`, and with `rescue_from Exception` that surfaces as a generic error). This PR restores them.

**SVG needs the block partially lifted.** librsvg is one of the untrusted loaders, so the initializer re-enables `VipsForeignLoadSvg{File,Buffer,Source}` after blocking — everything else (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm, FITS, ImageMagick delegates) stays blocked. Order matters: the unblock must run *after* `image_processing/vips` is loaded, or its own `block_untrusted` call undoes it. The initializer therefore requires it explicitly, and `spec/lib/vips_blocking_spec.rb` guards that trap.

Residual risk: librsvg stays reachable, but only for content marcel already sniffed as SVG. To drop SVG entirely, delete that block in the initializer and remove `image/svg+xml` from both uploaders.

## Verified

- `bundle exec rspec spec/lib/vips_blocking_spec.rb spec/uploaders/image_uploaders_spec.rb` → 13 examples, 0 failures. Rubocop clean on the added files (the pre-existing offenses in the two uploaders are untouched).
- Covered: Netpbm and other untrusted loaders blocked; PNG/JPEG/SVG still load; SVG still processes through an `ImageProcessing::Vips` pipeline; SVG survives the uploader require that used to re-block it; PDF bytes renamed `.png` are rejected by the allowlist.

## Worth a look in review

- The allowlist newly rejects HEIC, WebP, GIF and TIFF, which upload fine today (their libvips loaders are not untrusted — checked). If clients send iPhone photos as HEIC for user avatars, add `image/heic` (and probably `image/webp`).
- Not addressed: neither uploader has a size limit, and there is no `rescue_from Vips::Error`, so a blocked format that slips past validation lands in `general_error` rather than a 422.
- IMB still ships rails 8.1.3; the 8.1.3.1 patch is unrelated to this fix but worth taking separately.

### DIFF
--- a/app/mailers/devise_mailer.rb
+++ b/app/mailers/devise_mailer.rb
@@ -1,5 +1,15 @@
+require 'marcel'
+
 class DeviseMailer < Devise::Mailer
   layout 'mailer'
+
+  # app.config.mailer.logo_b64 is client-settable (Api::V1::AppAdminController
+  # permits it) and #logo hands it straight to libvips with no uploader in front,
+  # so it needs its own content check: config/initializers/vips.rb re-enables the
+  # SVG loader process-wide for the image uploaders, and without this gate an app
+  # admin could store an SVG that librsvg parses during mail rendering.
+  # Both AppAdminSerializer and AppInfoSerializer document the field as a PNG.
+  LOGO_MIME_TYPES = %w[image/png image/jpeg].freeze
 
   def unlock_instructions(user, token, opts = {})
     headers ApplicationMailer.default_headers
@@ -206,6 +216,18 @@ class DeviseMailer < Devise::Mailer
       _tmp.binmode
       _tmp.write _file.read
       _tmp.try :rewind
+
+      # Sniffed, not the content type Base64StringIO carried over from the data
+      # URI — that one is whatever the client declared.
+      _mime_type = Marcel::MimeType.for(_tmp)
+      unless LOGO_MIME_TYPES.include?(_mime_type)
+        Rails.logger.warn("DeviseMailer: ignoring mail logo of type #{_mime_type.inspect}")
+        _tmp.close
+        _tmp.delete
+        return nil
+      end
+      _tmp.try :rewind
+
       _logo = Vips::Image.new_from_file _tmp.path
       attachments.inline['logo.png'] = _tmp.read
       _tmp.close
--- a/app/mailers/devise_mailer.rb
+++ b/app/mailers/devise_mailer.rb
@@ -8,8 +8,19 @@ class DeviseMailer < Devise::Mailer
   # so it needs its own content check: config/initializers/vips.rb re-enables the
   # SVG loader process-wide for the image uploaders, and without this gate an app
   # admin could store an SVG that librsvg parses during mail rendering.
-  # Both AppAdminSerializer and AppInfoSerializer document the field as a PNG.
-  LOGO_MIME_TYPES = %w[image/png image/jpeg].freeze
+  #
+  # Everything libvips reads with a fuzzed (trusted) loader, i.e. the uploader
+  # allowlist minus SVG — librsvg is exactly the loader that must not be reachable
+  # from here. Deliberately not narrower than that: until now the only gate on this
+  # path was libvips itself, so GIF, TIFF, WebP and HEIC logos rendered, and
+  # rejecting them here would make a working logo vanish from every mail with
+  # nothing but a log line to say why. The serializers document the field as a PNG,
+  # but nothing has ever enforced that — Actors::App::Config::Mailer#logo_b64=
+  # stores whatever the admin picked in the UI.
+  #
+  # Pre-existing quirk this does not address: attachments.inline hardcodes
+  # 'logo.png', so a JPEG or GIF logo is declared to the client as image/png.
+  LOGO_MIME_TYPES = (ImageUploader::MIME_TYPES - ['image/svg+xml']).freeze
 
   def unlock_instructions(user, token, opts = {})
     headers ApplicationMailer.default_headers
@@ -204,45 +215,23 @@ class DeviseMailer < Devise::Mailer
     @app ||= Actors::App.named(app_context).first
   end
 
+  # defined?, not ||=: nil is a meaningful result here (no logo configured, or one
+  # LOGO_MIME_TYPES rejects), and prepare_logo -> logo_size ->
+  # logo_width/height/style each re-enter this method. Without memoising nil, one
+  # rejected logo cost five base64 decodes of up to 1 MB, five Tempfiles, five
+  # marcel sniffs and five identical warnings on every mail that app sends.
   def logo
-    @logo ||= begin
-      _logo_b64 = app.config.try(:mailer).try(:logo_b64) rescue nil
-      return nil unless _logo_b64.present?
+    return @logo if defined?(@logo)
 
-      _file = Base64StringIO.from_base64(_logo_b64, 'logo')
-      _file.try :rewind
-      _file.binmode
-      _tmp = Tempfile.new(_file.original_filename)
-      _tmp.binmode
-      _tmp.write _file.read
-      _tmp.try :rewind
-
-      # Sniffed, not the content type Base64StringIO carried over from the data
-      # URI — that one is whatever the client declared.
-      _mime_type = Marcel::MimeType.for(_tmp)
-      unless LOGO_MIME_TYPES.include?(_mime_type)
-        Rails.logger.warn("DeviseMailer: ignoring mail logo of type #{_mime_type.inspect}")
-        _tmp.close
-        _tmp.delete
-        return nil
-      end
-      _tmp.try :rewind
-
-      _logo = Vips::Image.new_from_file _tmp.path
-      attachments.inline['logo.png'] = _tmp.read
-      _tmp.close
-      _tmp.delete
-      _logo
-    end
+    @logo = build_logo
   end
 
   def logo_size
-    @logo_size ||= begin
-      width = 250
-      _logo = logo
-      return "" if _logo.nil?
-      "#{width}x#{(_logo.height / (_logo.width.to_f / width)).to_i}"
-    end
+    return @logo_size if defined?(@logo_size) # same reason as #logo
+
+    width = 250
+    _logo = logo
+    @logo_size = _logo.nil? ? '' : "#{width}x#{(_logo.height / (_logo.width.to_f / width)).to_i}"
   end
 
   def logo_width
@@ -262,6 +251,36 @@ class DeviseMailer < Devise::Mailer
       _width, _height = logo_size.split('x')
       "width:#{_width}px; height:#{_height}px;"
     end
+  end
+
+  def build_logo
+    _logo_b64 = app.config.try(:mailer).try(:logo_b64) rescue nil
+    return nil unless _logo_b64.present?
+
+    _file = Base64StringIO.from_base64(_logo_b64, 'logo')
+    _file.try :rewind
+    _file.binmode
+    _tmp = Tempfile.new(_file.original_filename)
+    _tmp.binmode
+    _tmp.write _file.read
+    _tmp.try :rewind
+
+    # Sniffed, not the content type Base64StringIO carried over from the data
+    # URI — that one is whatever the client declared.
+    _mime_type = Marcel::MimeType.for(_tmp)
+    unless LOGO_MIME_TYPES.include?(_mime_type)
+      Rails.logger.warn("DeviseMailer: ignoring mail logo of type #{_mime_type.inspect}")
+      _tmp.close
+      _tmp.delete
+      return nil
+    end
+    _tmp.try :rewind
+
+    _logo = Vips::Image.new_from_file _tmp.path
+    attachments.inline['logo.png'] = _tmp.read
+    _tmp.close
+    _tmp.delete
+    _logo
   end
 
   def prepare_logo
--- a/app/uploaders/actor_image_uploader.rb
+++ b/app/uploaders/actor_image_uploader.rb
@@ -11,11 +11,12 @@ class ActorImageUploader < Shrine
   plugin :validation_helpers
 
   Attacher.validate do
-    # Only formats we actually render derivatives for. Without this, arbitrary
-    # uploaded bytes reach libvips and the loader is picked by content sniffing —
-    # see config/initializers/vips.rb (CVE-2026-66066). The MIME type comes from
+    # Same allowlist as ImageUploader, deliberately shared rather than copied so
+    # the two cannot drift: both feed the same libvips loaders, and the reasoning
+    # behind the list lives on ImageUploader::MIME_TYPES
+    # (config/initializers/vips.rb, CVE-2026-66066). The MIME type comes from
     # marcel content analysis, not from the client-supplied header.
-    validate_mime_type_inclusion %w[image/png image/jpeg image/svg+xml]
+    validate_mime_type_inclusion ImageUploader::MIME_TYPES
   end
 
   add_metadata do |io, context|
--- a/app/uploaders/actor_image_uploader.rb
+++ b/app/uploaders/actor_image_uploader.rb
@@ -8,6 +8,15 @@ class ActorImageUploader < Shrine
   plugin :add_metadata
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
+  plugin :validation_helpers
+
+  Attacher.validate do
+    # Only formats we actually render derivatives for. Without this, arbitrary
+    # uploaded bytes reach libvips and the loader is picked by content sniffing —
+    # see config/initializers/vips.rb (CVE-2026-66066). The MIME type comes from
+    # marcel content analysis, not from the client-supplied header.
+    validate_mime_type_inclusion %w[image/png image/jpeg image/svg+xml]
+  end
 
   add_metadata do |io, context|
     filename = context[:record].id
--- a/app/uploaders/actor_image_uploader.rb
+++ b/app/uploaders/actor_image_uploader.rb
@@ -9,6 +9,9 @@ class ActorImageUploader < Shrine
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
   plugin :validation_helpers
+  # See ImageUploader: `model, cache: false` uploads to :store before validating,
+  # so rejected files need destroying or they stay in the bucket unreferenced.
+  plugin :remove_invalid
 
   Attacher.validate do
     # Same allowlist as ImageUploader, deliberately shared rather than copied so
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -8,6 +8,15 @@ class ImageUploader < Shrine
   plugin :add_metadata
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
+  plugin :validation_helpers
+
+  Attacher.validate do
+    # Only formats we actually render derivatives for. Without this, arbitrary
+    # uploaded bytes reach libvips and the loader is picked by content sniffing —
+    # see config/initializers/vips.rb (CVE-2026-66066). The MIME type comes from
+    # marcel content analysis, not from the client-supplied header.
+    validate_mime_type_inclusion %w[image/png image/jpeg image/svg+xml]
+  end
 
   add_metadata do |io, context|
     filename = context[:record].id
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -10,12 +10,34 @@ class ImageUploader < Shrine
   plugin :derivatives, versions_compatibility: true
   plugin :validation_helpers
 
+  # Formats libvips can read with a fuzzed (trusted) loader, plus SVG, which
+  # config/initializers/vips.rb re-enables deliberately. Without this list,
+  # arbitrary uploaded bytes reach libvips and the loader is picked by content
+  # sniffing — see that initializer and CVE-2026-66066. None of these loaders is
+  # part of libvips' untrusted set, so allowing them costs nothing against the
+  # CVE; what the list buys is rejecting content that is not an image at all,
+  # with a clean validation error. HEIC/HEIF/AVIF and WebP matter because avatars
+  # come off phones, GIF and TIFF because they were uploadable before this list
+  # existed and turning a working upload into an error is the worse failure.
+  #
+  # The MIME type comes from marcel content analysis (determine_mime_type
+  # plugin), not from the client-supplied header — which matters for
+  # User#image_b64= and Actor#image_b64=, where Base64StringIO carries the
+  # content type the client declared in the data URI.
+  MIME_TYPES = %w[
+    image/png
+    image/jpeg
+    image/svg+xml
+    image/heic
+    image/heif
+    image/avif
+    image/webp
+    image/gif
+    image/tiff
+  ].freeze
+
   Attacher.validate do
-    # Only formats we actually render derivatives for. Without this, arbitrary
-    # uploaded bytes reach libvips and the loader is picked by content sniffing —
-    # see config/initializers/vips.rb (CVE-2026-66066). The MIME type comes from
-    # marcel content analysis, not from the client-supplied header.
-    validate_mime_type_inclusion %w[image/png image/jpeg image/svg+xml]
+    validate_mime_type_inclusion ImageUploader::MIME_TYPES
   end
 
   add_metadata do |io, context|
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -12,6 +12,13 @@ class ImageUploader < Shrine
   # With `model, cache: false` an assignment uploads to permanent :store before
   # the validation below runs, so a rejected file would sit in the bucket
   # unreferenced forever — the record never saves. remove_invalid destroys it.
+  # Precisely: it reverts the *attacher* (record.image is the previous file again,
+  # the rejected object is deleted) but never writes through to the model, because
+  # deassign uses load_data rather than set. So record.image_data still holds the
+  # deleted file's JSON in memory. Harmless while the attacher's errors keep the
+  # record from saving — the trap is a save(validate: false) on the same in-memory
+  # record, e.g. User#tenant_access_group_ids, which would persist a reference to a
+  # deleted key.
   plugin :remove_invalid
 
   # Formats libvips can read with a fuzzed (trusted) loader, plus SVG, which
--- a/app/uploaders/image_uploader.rb
+++ b/app/uploaders/image_uploader.rb
@@ -9,6 +9,10 @@ class ImageUploader < Shrine
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
   plugin :validation_helpers
+  # With `model, cache: false` an assignment uploads to permanent :store before
+  # the validation below runs, so a rejected file would sit in the bucket
+  # unreferenced forever — the record never saves. remove_invalid destroys it.
+  plugin :remove_invalid
 
   # Formats libvips can read with a fuzzed (trusted) loader, plus SVG, which
   # config/initializers/vips.rb re-enables deliberately. Without this list,
--- a/config/initializers/vips.rb
+++ b/config/initializers/vips.rb
@@ -29,10 +29,14 @@ Vips.block_untrusted(true)
 # every other untrusted loader/saver (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm,
 # FITS, Matlab, OpenSlide, anything delegated to ImageMagick) stays blocked.
 #
-# ImageUploader and ActorImageUploader validate against a png/jpeg/svg allowlist
-# with content sniffing (marcel) before any of this runs, so only content that
-# actually looks like SVG reaches librsvg. To drop SVG support entirely, delete
-# this block and remove image/svg+xml from both uploaders.
+# ImageUploader and ActorImageUploader both validate against
+# ImageUploader::MIME_TYPES with content sniffing (marcel) before any of this
+# runs, so only content that actually looks like SVG reaches librsvg — including
+# on the User#image_b64= / Actor#image_b64= paths, where the content type the
+# client declared in the data URI is not what gets validated.
+#
+# To drop SVG support entirely, delete this block and remove image/svg+xml from
+# ImageUploader::MIME_TYPES.
 %w[
   VipsForeignLoadSvgFile
   VipsForeignLoadSvgBuffer
--- a/config/initializers/vips.rb
+++ b/config/initializers/vips.rb
@@ -1,0 +1,40 @@
+# libvips reads and writes formats through "loaders" and "savers", some of which
+# it flags as "unfuzzed"/untrusted: they are backed by third-party libraries that
+# are only safe for trusted content. Feeding attacker-supplied files to them can
+# disclose arbitrary files readable by this process, potentially RCE.
+# See CVE-2026-66066 / GHSA-xr9x-r78c-5hrm — reported against Active Storage,
+# which we do not load (see config/application.rb); the underlying problem is
+# libvips, which we do use.
+#
+# image_processing 2.0 already calls Vips.block_untrusted(true) when
+# image_processing/vips is required, but not every path into libvips goes through
+# that gem: DeviseMailer#logo hands the app's configured logo_b64 straight to
+# Vips::Image.new_from_file. Blocking here makes it explicit and independent of
+# load order.
+#
+# Requiring image_processing/vips first is deliberate: that require calls
+# block_untrusted itself, which re-blocks the SVG loaders re-enabled below. The
+# require is idempotent, so the uploaders' own requires are no-ops and cannot
+# undo this.
+require 'image_processing/vips'
+
+unless Vips.at_least_libvips?(8, 13)
+  raise "libvips #{Vips.version_string} cannot block untrusted operations — 8.13 or newer is required"
+end
+
+Vips.block_untrusted(true)
+
+# The SVG loader (librsvg) is one of the untrusted ones, but SVG is an accepted
+# upload format for user and actor images. Re-enable just those three loaders;
+# every other untrusted loader/saver (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm,
+# FITS, Matlab, OpenSlide, anything delegated to ImageMagick) stays blocked.
+#
+# ImageUploader and ActorImageUploader validate against a png/jpeg/svg allowlist
+# with content sniffing (marcel) before any of this runs, so only content that
+# actually looks like SVG reaches librsvg. To drop SVG support entirely, delete
+# this block and remove image/svg+xml from both uploaders.
+%w[
+  VipsForeignLoadSvgFile
+  VipsForeignLoadSvgBuffer
+  VipsForeignLoadSvgSource
+].each { |loader| Vips.block(loader, false) }
--- a/config/initializers/vips.rb
+++ b/config/initializers/vips.rb
@@ -25,9 +25,16 @@ end
 Vips.block_untrusted(true)
 
 # The SVG loader (librsvg) is one of the untrusted ones, but SVG is an accepted
-# upload format for user and actor images. Re-enable just those three loaders;
-# every other untrusted loader/saver (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm,
-# FITS, Matlab, OpenSlide, anything delegated to ImageMagick) stays blocked.
+# upload format for user and actor images. Re-enable it; every other untrusted
+# loader/saver (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm, FITS, Matlab,
+# OpenSlide, anything delegated to ImageMagick) stays blocked. Naming the parent
+# class covers the file, buffer and source variants — vips_operation_block_set
+# applies to the named class and everything below it.
+#
+# This is process-wide, not per-uploader, so every caller that reaches libvips
+# needs its own content check, not just the uploaders — see
+# DeviseMailer::LOGO_MIME_TYPES for the mail logo, which has no uploader in front
+# of it.
 #
 # ImageUploader and ActorImageUploader both validate against
 # ImageUploader::MIME_TYPES with content sniffing (marcel) before any of this
@@ -37,8 +44,7 @@ Vips.block_untrusted(true)
 #
 # To drop SVG support entirely, delete this block and remove image/svg+xml from
 # ImageUploader::MIME_TYPES.
-%w[
-  VipsForeignLoadSvgFile
-  VipsForeignLoadSvgBuffer
-  VipsForeignLoadSvgSource
-].each { |loader| Vips.block(loader, false) }
+# NOTE: Vips.block is a no-op for an unknown operation name, so a typo or a
+# libvips built without librsvg fails silently — SVG simply stays blocked.
+# spec/lib/vips_blocking_spec.rb is what catches that.
+Vips.block('VipsForeignLoadSvg', false)
--- a/spec/lib/vips_blocking_spec.rb
+++ b/spec/lib/vips_blocking_spec.rb
@@ -1,0 +1,56 @@
+require 'rails_helper'
+
+# Guards config/initializers/vips.rb (CVE-2026-66066 / GHSA-xr9x-r78c-5hrm).
+#
+# The load order is the fragile part: image_processing/vips calls
+# Vips.block_untrusted(true) whenever it is required, which re-blocks the SVG
+# loaders the initializer re-enables. The initializer therefore requires
+# image_processing/vips itself, before unblocking SVG. Touching the uploaders
+# here re-triggers that require the way eager loading does in production.
+describe 'libvips untrusted operation blocking', type: :model do
+  def svg(size)
+    %(<svg xmlns="http://www.w3.org/2000/svg" width="#{size}" height="#{size}">) +
+      %(<rect width="#{size}" height="#{size}"/></svg>)
+  end
+
+  def tempfile(bytes, extension)
+    file = Tempfile.new(['vips-blocking', extension])
+    file.binmode
+    file.write(bytes)
+    file.rewind
+    file
+  end
+
+  def loadable?(file)
+    Vips::Image.new_from_file(file.path).avg
+    true
+  rescue Vips::Error
+    false
+  end
+
+  before { [ImageUploader, ActorImageUploader].each(&:name) }
+
+  it 'blocks untrusted loaders' do
+    # Netpbm stands in for the whole untrusted set (BMP, ICO, PSD, JPEG XL,
+    # JPEG 2000, FITS, Matlab, OpenSlide, ImageMagick delegates).
+    expect(loadable?(tempfile("P3\n1 1\n255\n0 0 0\n", '.ppm'))).to be false
+  end
+
+  it 'still allows the SVG loader, which user and actor images accept' do
+    expect(loadable?(tempfile(svg(4), '.svg'))).to be true
+  end
+
+  it 'still allows PNG' do
+    expect(loadable?(tempfile(Vips::Image.black(4, 4).write_to_buffer('.png'), '.png'))).to be true
+  end
+
+  it 'still allows JPEG' do
+    expect(loadable?(tempfile(Vips::Image.black(4, 4).write_to_buffer('.jpg'), '.jpg'))).to be true
+  end
+
+  it 'processes an SVG through the uploader pipeline' do
+    source = tempfile(svg(40), '.svg').path
+    processed = ImageProcessing::Vips.source(source).resize_to_limit(20, 20).convert('png').call
+    expect(File.size(processed.path)).to be > 0
+  end
+end
--- a/spec/lib/vips_blocking_spec.rb
+++ b/spec/lib/vips_blocking_spec.rb
@@ -32,7 +32,21 @@ describe 'libvips untrusted operation blocking', type: :model do
   it 'blocks untrusted loaders' do
     # Netpbm stands in for the whole untrusted set (BMP, ICO, PSD, JPEG XL,
     # JPEG 2000, FITS, Matlab, OpenSlide, ImageMagick delegates).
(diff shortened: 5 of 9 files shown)


config/initializers/vips.rb
```diff
@@ -0,0 +1,40 @@
...
+Vips.block_untrusted(true)
+
+# The SVG loader (librsvg) is one of the untrusted ones, but SVG is an accepted
+# upload format for user and actor images. Re-enable just those three loaders;
+# every other untrusted loader/saver (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm,
+# FITS, Matlab, OpenSlide, anything delegated to ImageMagick) stays blocked.
+#
+# ImageUploader and ActorImageUploader validate against a png/jpeg/svg allowlist
+# with content sniffing (marcel) before any of this runs, so only content that
+# actually looks like SVG reaches librsvg. To drop SVG support entirely, delete
```
SCjona: This claim does not hold for the path the initializer was written for.

`Vips.block` flips a global flag on the operation class (`vips_operation_block_set`), so unblocking the SVG loaders unblocks them process-wide — including for `DeviseMailer#logo` (`app/mailers/devise_mailer.rb:209`), which hands `app.config.mailer.logo_b64` straight to `Vips::Image.new_from_file` with no MIME check at all. That value is client-settable: `Api::V1::AppAdminController::PERMIT_UPDATE` permits `config.mailer.logo_b64`.

Net effect of this PR on that path is a regression, not a fix: today `image_processing/vips` blocks svgload for the mailer too, so an SVG logo simply fails. After this PR an app admin can store an SVG that librsvg parses inside mail rendering. Admin-scoped, so not urgent — but it is exactly the gap the header comment says the initializer closes, and the uploader allowlist provides no coverage there.

Two options:

1. Validate in the mailer before handing bytes to libvips — `Marcel::MimeType.for(_tmp)` against `%w[image/png image/jpeg]`, return `nil` otherwise. Both serializers already document this field as "A PNG logo as mail header in Base64 encoded format" (`app_admin_serializer.rb:137`, `app_info_serializer.rb:107`), so this only enforces the documented contract, and `logo`/`logo_size` already handle `nil`.
2. If you would rather not touch the mailer in this PR, at minimum reword this comment — as written it tells the next reader that librsvg is unreachable without marcel in front of it, which is the assumption that makes the mailer gap easy to miss.

Minor while you are here: `Vips.block` is documented as a no-op when the operation name is not found, so a typo or a libvips build without librsvg fails silently (SVG stays blocked, uploads 500 at derivative time). `spec/lib/vips_blocking_spec.rb` catches that in CI, which is the main thing. You could also collapse the three entries to `Vips.block("VipsForeignLoadSvg", false)` — `vips_operation_block_set` applies to the named class and everything below it.

app/uploaders/image_uploader.rb
```diff
@@ -8,6 +8,15 @@ class ImageUploader < Shrine
...
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
+  plugin :validation_helpers
+
+  Attacher.validate do
+    # Only formats we actually render derivatives for. Without this, arbitrary
+    # uploaded bytes reach libvips and the loader is picked by content sniffing —
+    # see config/initializers/vips.rb (CVE-2026-66066). The MIME type comes from
+    # marcel content analysis, not from the client-supplied header.
+    validate_mime_type_inclusion %w[image/png image/jpeg image/svg+xml]
```
SCjona: The allowlist keeps rejected bytes away from libvips, but not out of the S3 bucket — and this PR is what makes them orphans.

With `plugin :model, cache: false` (line 6), `record.image = file` goes through `Attacher#model_assign` -> `attach` (shrine `model.rb:118`), and `attach` uploads *first*, then calls `change`, which is where the validation plugin hooks in (shrine `attacher.rb:116-120`). So the sequence is: upload to permanent `:store`, then validate, then add the error. The record never saves, so nothing ever references or deletes that object.

Before this PR there were no validations, so every uploaded file ended up attached to a record. Now any authenticated user can push arbitrary content of any MIME type into permanent storage and have it stay there unreferenced — and with no size limit either, that is a cheap way to fill the bucket.

One line fixes it:

```ruby
plugin :validation_helpers
plugin :remove_invalid
```

`remove_invalid` wraps `validate` and calls `destroy` on the attached file when `errors.any?` (`shrine/plugins/remove_invalid.rb`). Same applies to `ActorImageUploader`.

Worth noting because it explains why the specs did not surface this: `spec/uploaders/image_uploaders_spec.rb` calls `attacher.assign(file)` directly, which routes through `attach_cached` to the `:cache` storage. The production path is `record.image =`, i.e. `model_assign`, which with `cache: false` goes to `:store`. Asserting on `Shrine.storages[:store].store.keys` after a rejected assign via `record.image =` would pin the behaviour down.

app/uploaders/actor_image_uploader.rb
```diff
@@ -8,6 +8,15 @@ class ActorImageUploader < Shrine
...
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
+  plugin :validation_helpers
+
+  Attacher.validate do
+    # Only formats we actually render derivatives for. Without this, arbitrary
+    # uploaded bytes reach libvips and the loader is picked by content sniffing —
+    # see config/initializers/vips.rb (CVE-2026-66066). The MIME type comes from
+    # marcel content analysis, not from the client-supplied header.
+    validate_mime_type_inclusion %w[image/png image/jpeg image/svg+xml]
```
SCjona: Byte-for-byte identical to `ImageUploader` — the whole file is, not just this block, and now the security-relevant allowlist is duplicated too. The next person to add `image/webp` will update one and not the other, and nothing fails loudly when that happens.

Not blocking, and I would not restructure both uploaders inside a security fix. But if you keep them separate, a `# keep in sync with ImageUploader` marker on both `Attacher.validate` blocks is cheap insurance. The alternative (`class ActorImageUploader < ImageUploader`) inherits the validate block automatically, since `Attacher.validate` defines a private `validate_block` method that a subclass attacher inherits — the only real difference between the two files is `resize_to_fill!` vs `resize_to_limit` in the derivatives.

SCjona (changes requested):
The core of this is right, and I verified the parts that are easy to get wrong:

- **Load order holds.** `image_processing/vips` 2.0.2 calls `Vips.block_untrusted(true)` at require time (`lib/image_processing/vips.rb:8`, guarded by `respond_to?` + `VIPS_BLOCK_UNTRUSTED`), so requiring it *before* the unblock is exactly the right sequencing, and the uploaders' own `require` is a no-op. Initializers run before eager loading, and `shrine.rb` sorts before `vips.rb`, so nothing beats it to the punch.
- **The API exists and is guarded correctly.** In ruby-vips 2.3.0 both `block_untrusted` and `block` are defined inside `if at_least_libvips?(8, 13)` (`lib/vips.rb:827`), so the boot guard is not cosmetic — without it you'd get `NoMethodError` instead of a clear message. `Vips.version_string` is attached (`lib/vips.rb:817`). `VipsForeignLoadSvg{File,Buffer,Source}` match the libvips GTypes.
- **The allowlist really is content-based.** Shrine's marcel analyzer passes `name: nil` unless `filename_fallback` is set (`determine_mime_type.rb:154`), so detection is pure magic-byte and the client-supplied filename/data-URI content type genuinely cannot influence it. The comment in both uploaders is accurate.
- **Your format claim checks out.** I grepped libvips master: `VIPS_OPERATION_UNTRUSTED` is present in `svgload.c`, `jxlload.c`, `jp2kload.c`, and absent from `heifload.c`, `webpload.c`, `gifload.c`, `tiffload.c`.

Two things I'd like resolved before merge, both concrete:

**1. The mailer logo path — the one this initializer exists for — now reaches librsvg unvalidated.** Inline on `config/initializers/vips.rb`.

**2. Rejected uploads are still written to permanent storage.** Inline on `image_uploader.rb`. One-line fix.

On your open questions:

- **HEIC/WebP/GIF/TIFF:** per the grep above, none of their loaders is untrusted, so adding them costs nothing against this CVE and the derivatives pipeline flattens everything to PNG anyway. If iOS clients send HEIC for avatars this will start 422-ing them, and the failure is invisible until a user complains. I'd add `image/webp` and `image/gif` unconditionally (browsers produce both), and `image/heic`/`image/heif` if you can confirm the deployed `libvips-dev` is built with libheif — a MIME type on the allowlist that libvips then can't load just moves the failure downstream.
- **`validate_max_size`:** agreed it's a separate follow-up. With the untrusted loaders blocked, a decompression bomb through the trusted PNG loader — or an SVG with a huge `width`/`height`, which librsvg will happily rasterize — is what's left. Same conclusion I reached on samedis-care-backend#4108.
- **`rescue_from Vips::Error`:** worth noting the validation failure itself *does* surface as a 422 — shrine-mongoid copies `attacher.errors` onto `record.errors[:image]`, so the `general_error` case only applies to bytes that pass the allowlist and then fail inside libvips.

Residual risk I'm accepting, same as on #4108: librsvg stays reachable for anything marcel sniffs as SVG. Documenting it in the initializer is the right call.

app/uploaders/image_uploader.rb
```diff
@@ -8,6 +8,37 @@ class ImageUploader < Shrine
   plugin :add_metadata
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
+  plugin :validation_helpers
```
SCjona, line 11: Still open from the previous review — the allowlist keeps rejected bytes away from libvips, but not out of the S3 bucket.

Re-verified against the shrine 3.8.0 in `Gemfile.lock`:

- `plugin :model, cache: false` (line 6) → `record.image = file` calls `Attacher#model_assign`, which with `model_cache? == false` calls `attach`, not `assign` (`plugins/model.rb:117-123`).
- `attach` does `file = upload(io, storage, **) if io` **then** `change(file)` (`attacher.rb:116-120`) — upload first, to `store_key`.
- The validation plugin's `attach` runs `super` first and only then `validation(validate)` (`plugins/validation.rb:39-43`).

So: uploaded to permanent `:store`, then validated, then the error is added. The record never saves, so nothing references or deletes that object. Before this PR there were no validations and every upload ended up attached; this PR is what creates the orphan class. Combined with no size limit, any authenticated user can push arbitrary content of any MIME type into the bucket and have it stay there.

```ruby
plugin :validation_helpers
plugin :remove_invalid
```

`remove_invalid` wraps `validate` and calls `destroy` in an `ensure` when `errors.any?`, then restores the previous data (`plugins/remove_invalid.rb`). Errors survive the `ensure`, so the 422 via shrine-mongoid is unaffected. Needs the same line in `ActorImageUploader`.

Worth pinning, and you already have the example that does it: `spec/uploaders/image_uploaders_spec.rb:119` (`record.image_b64 = ...`) is the only spec that goes through the production `record.image =` → `model_assign` → `attach` path — every other one calls `attacher.assign`, which routes to `:cache`. Adding `expect(ImageUploader.storages[:store].store).to be_empty` to that example pins the behaviour with no new fixture.

config/initializers/vips.rb
```diff
@@ -0,0 +1,44 @@
...
+
+Vips.block_untrusted(true)
+
+# The SVG loader (librsvg) is one of the untrusted ones, but SVG is an accepted
+# upload format for user and actor images. Re-enable just those three loaders;
+# every other untrusted loader/saver (BMP, ICO, PSD, JPEG XL, JPEG 2000, Netpbm,
+# FITS, Matlab, OpenSlide, anything delegated to ImageMagick) stays blocked.
+#
+# ImageUploader and ActorImageUploader both validate against
+# ImageUploader::MIME_TYPES with content sniffing (marcel) before any of this
```
SCjona, line 40: Still open from the previous review. `Vips.block` flips a global flag on the operation class (`vips_operation_block_set`), so unblocking the SVG loaders unblocks them process-wide — including for `DeviseMailer#logo` (`app/mailers/devise_mailer.rb:209`), which hands `app.config.mailer.logo_b64` straight to `Vips::Image.new_from_file` with no MIME check. I re-confirmed the path is client-settable: `Api::V1::AppAdminController::PERMIT_UPDATE` permits `config.mailer.logo_b64` (`app_admin_controller.rb:25`), and `Actors::App::Config::Mailer#logo_b64=` (`app/models/actors/app.rb:97`) runs `ensure_base64`, which by its own comment accepts form-encoded uploads too. `Vips` appears nowhere else outside the uploaders — the mailer is the *only* direct caller, and it is the one the header comment (lines 9-13) names as the reason this initializer exists.

Net effect on that path is a regression: today `image_processing/vips`'s own `block_untrusted` blocks svgload for the mailer too, so an SVG logo just fails. After this PR an app admin can store an SVG that librsvg parses inside mail rendering. Admin-scoped, so not urgent — but this file is what opens it, and the uploader allowlist gives no coverage there.

These lines make that easy to miss: "*ImageUploader and ActorImageUploader both validate ... before any of this runs, so only content that actually looks like SVG reaches librsvg*". True for the two upload paths, false for the mailer, and the sentence reads as a claim about librsvg in general.

Two options, unchanged from last round:

1. Validate in the mailer before handing bytes to libvips — `Marcel::MimeType.for(_tmp)` against `%w[image/png image/jpeg]`, return `nil` otherwise. Both serializers already document the field as "A PNG logo as mail header in Base64 encoded format" (`app_admin_serializer.rb:137`, `app_info_serializer.rb:107`), so that only enforces the documented contract, and `logo`/`logo_size` already handle `nil`.
2. If you'd rather not touch the mailer here, reword this paragraph so it scopes the claim to the uploaders and names the mailer as an uncovered caller. Then it's a documented gap instead of a silent one.

Minor, unchanged: `Vips.block` is a no-op for an unknown operation name, so a typo or a build without librsvg fails silently. `spec/lib/vips_blocking_spec.rb` catches that in CI, which is the main thing. `Vips.block("VipsForeignLoadSvg", false)` would collapse the three entries — `vips_operation_block_set` applies to the named class and everything below it.

app/uploaders/actor_image_uploader.rb
```diff
@@ -8,6 +8,16 @@ class ActorImageUploader < Shrine
   plugin :add_metadata
   plugin :determine_mime_type, analyzer: :marcel
   plugin :derivatives, versions_compatibility: true
+  plugin :validation_helpers
```
SCjona, line 11: Same one-liner as on `ImageUploader` — `plugin :remove_invalid` alongside `plugin :validation_helpers`. `ActorImageUploader` has the identical `plugin :model, cache: false`, so rejected actor images are written to `:store` and orphaned the same way. See the `image_uploader.rb` comment for the trace.

spec/lib/vips_blocking_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    Vips::Image.new_from_file(file.path).avg
+    true
+  rescue Vips::Error
+    false
+  end
+
+  it 'blocks untrusted loaders' do
+    # Netpbm stands in for the whole untrusted set (BMP, ICO, PSD, JPEG XL,
+    # JPEG 2000, FITS, Matlab, OpenSlide, ImageMagick delegates).
+    expect(loadable?(tempfile("P3\n1 1\n255\n0 0 0\n", '.ppm'))).to be false
```
SCjona: Non-blocking, but this is the example carrying the central security assertion and it can pass for the wrong reason. `loadable?` swallows every `Vips::Error` alike, and libvips reports a blocked loader through the same "not a known file format" path as a loader that simply isn't compiled into the build — so on an image without ppmload this is green whether or not `block_untrusted` did anything.

A positive control makes it real:

```ruby
it 'blocks untrusted loaders' do
  file = tempfile("P3\n1 1\n255\n0 0 0\n", '.ppm')
  begin
    Vips.block('VipsForeignLoadPpmFile', false)
    expect(loadable?(file)).to be(true), 'ppmload is not in this libvips build; the block assertion below proves nothing'
  ensure
    Vips.block('VipsForeignLoadPpmFile', true)
  end
  expect(loadable?(file)).to be false
end
```

Same shape applies to the `.heic`/`.webp` example below, which `next`s past both extensions and passes vacuously if neither saver exists — a `skip` there would at least make the gap visible in the run output.

SCjona (changes requested):
Re-reviewed at `2d9407c`. The three items you carried over from samedis-care-backend#4108 all check out:

- `MIME_TYPES` now covers HEIC/HEIF/AVIF/WebP/GIF/TIFF, and `ActorImageUploader` references the constant rather than copying it — that resolves my drift comment. Lazy constant lookup inside the `Attacher.validate` block is fine under Zeitwerk.
- The storage override is now on the two classes under test, which is the right fix for `shrine.rb#inherited` duping `@storages` at class-definition time.
- The `Tempfile` in the pipeline example is referenced for the duration of `#call`, and dropping the `before` hook is correct — the require is already in `$LOADED_FEATURES` after boot.

The two things I asked for before merge are still open on this head, though — I don't think either was intentional, since your update comment only speaks to the #4108 carry-over. Re-raising both with the verification I did this round.

**1. `remove_invalid` is still missing** — rejected bytes still land in permanent storage. Re-confirmed against shrine 3.8.0: `attach` (`attacher.rb:116-120`) uploads *before* `change`, and the validation plugin hooks the post-`super` path (`plugins/validation.rb:39-43`). With `plugin :model, cache: false`, `model_assign` routes to `attach`, i.e. straight to `:store`. Inline on `image_uploader.rb`; same one-liner needed on `ActorImageUploader`.

**2. The mailer logo path still reaches librsvg unvalidated**, and the initializer comment now asserts the opposite. Inline on `config/initializers/vips.rb`.

Also noted from your comment: the size limit and `rescue_from Vips::Error` are fine as follow-ups. One non-blocking observation on the blocking spec inline.

app/mailers/devise_mailer.rb
```diff
@@ -1,6 +1,16 @@
...
 class DeviseMailer < Devise::Mailer
   layout 'mailer'
 
+  # app.config.mailer.logo_b64 is client-settable (Api::V1::AppAdminController
+  # permits it) and #logo hands it straight to libvips with no uploader in front,
+  # so it needs its own content check: config/initializers/vips.rb re-enables the
+  # SVG loader process-wide for the image uploaders, and without this gate an app
+  # admin could store an SVG that librsvg parses during mail rendering.
+  # Both AppAdminSerializer and AppInfoSerializer document the field as a PNG.
+  LOGO_MIME_TYPES = %w[image/png image/jpeg].freeze
```
SCjona: This allowlist is narrower than the status quo on this path, and the failure is silent — worth a look at production data before merging.

Pre-PR the gate here was libvips itself: `Vips::Image.new_from_file` succeeded for anything whose loader was not blocked, so PNG, JPEG, **GIF**, TIFF, WebP and HEIC logos all rendered. Only the untrusted loaders (SVG, BMP, ICO) failed, and they failed loudly. So for those three this PR is a straight improvement (see the review body). For the rest it is a downgrade: the logo disappears from every mail that app sends, with nothing but a `logger.warn` to say so.

"The serializers document PNG" is also weaker than it looks as an argument that only PNGs are stored. `Actors::App::Config::Mailer#logo_b64=` (`app/models/actors/app.rb:97`) runs `ensure_base64`, which by its own comment converts form-encoded uploads — so whatever file an app admin picked in the UI is what got stored, and nothing has ever enforced the documented type. The only constraint today is `validates_length_of ... maximum: 1.megabyte` (`app.rb:43`).

GIF is the case I would actually worry about: mail clients render it, so it works for those tenants right now, and `image/gif` on the list preserves that at zero cost against the CVE (`gifload` is not in the untrusted set). TIFF/WebP/HEIC are more defensible to drop — but the reason is not the allowlist, it is that `attachments.inline['logo.png']` hardcodes the name, so Mail declares `image/png` regardless and those were already being mislabelled to the client. That is equally true of the `image/jpeg` you are keeping. Since you now have `_mime_type` in hand where the attachment is built, making the declared type match is nearly free if you want to close that too.

Either way, a console pass over the stored `config.mailer.logo_b64` prefixes across apps would tell you whether any of this is theoretical.

app/uploaders/image_uploader.rb
```diff
@@ -8,6 +8,41 @@ class ImageUploader < Shrine
...
+  # With `model, cache: false` an assignment uploads to permanent :store before
+  # the validation below runs, so a rejected file would sit in the bucket
+  # unreferenced forever — the record never saves. remove_invalid destroys it.
+  plugin :remove_invalid
```
SCjona, line 22: This does what the comment says — the rejected object is gone from `:store`, I traced it. One leftover it does not cover: `remove_invalid` restores the previous attachment on the *attacher*, but never writes it back to the model attribute.

- `deassign` → `load_data(@previous.data)` (`plugins/remove_invalid.rb`)
- `Attacher#load_data` only assigns `@file` (`attacher.rb:304`)
- the `model` plugin writes the column exclusively from `set` (`plugins/model.rb:120-127`), and `deassign` never goes through `set`

So after a rejected `record.image = ...`:

| | state |
|---|---|
| `:store` | object deleted ✓ |
| `attacher.file` / `record.image` | previous file restored ✓ |
| `record.image_data` | still the JSON of the file that was just deleted ✗ |

Shrine's own docs for the plugin only claim the reader reverts, and that part is true. Ordinarily the stale column is harmless, because the attacher errors make the record invalid and the save never happens.

What makes it worth writing down rather than ignoring is that this repo has validation-skipping saves on the same in-memory `User`: `User#tenant_access_group_ids` (`app/models/user.rb:766`) does `self.save(validate: false) if self.changes.any?`, and a dirty `image_data` is one of those changes. That persists a reference to a deleted key, and the avatar is then permanently broken with no way to notice.

To be straight about the likelihood: I could not construct a request that actually reaches it. Authorization — and therefore the access-group recompute — runs before the action rather than after the rejected assign, and `Api::V1::App::UserController#update` renders the 422 from `check_for_errors` and stops. So this is a latent hazard, not a live bug, and a follow-up is fine by me.

`remove_invalid` offers no hook, so closing it means overriding `Attacher#validate` to `write` when `errors.any?`. Mostly I want it recorded so the next person does not assume `record.image_data` is clean after a rejected assign — which is exactly the assumption that made the orphaned-object problem easy to miss the first time round.

app/mailers/devise_mailer.rb
```diff
@@ -206,6 +216,18 @@ def logo
...
       _tmp.try :rewind
+
+      # Sniffed, not the content type Base64StringIO carried over from the data
+      # URI — that one is whatever the client declared.
+      _mime_type = Marcel::MimeType.for(_tmp)
+      unless LOGO_MIME_TYPES.include?(_mime_type)
+        Rails.logger.warn("DeviseMailer: ignoring mail logo of type #{_mime_type.inspect}")
+        _tmp.close
+        _tmp.delete
+        return nil
```
SCjona: This `return` exits `logo` without assigning `@logo`, so the `||=` never memoizes the rejection. Same shape as the pre-existing `return nil unless _logo_b64.present?` above it, but the body it now skips is far more expensive than that one was.

`prepare_logo` calls `logo` and then `logo_size`; `logo_size` itself does `return "" if _logo.nil?` without assigning `@logo_size`, so `logo_width`, `logo_height` and `logo_style` each re-enter it, and each of those re-enters `logo`. That is 5 invocations per mail for one rejected logo:

- 5 × `Base64.decode64` of up to 1 MB (`validates_length_of :logo_b64, maximum: 1.megabyte`)
- 5 × Tempfile created, written, closed, unlinked
- 5 × `Marcel::MimeType.for`, which marcel's own docs describe as a slow operation (`by_magic` walks the magic table with seeks and reads)
- 5 × the identical `logger.warn`, for every mail that app sends

The log noise is the part that will actually mislead someone: a tenant with a rejected logo produces five warnings per password reset, and nothing in the output says it is one logo rather than five.

Cheap fix, since `nil` is a meaningful memoized value here:

```ruby
def logo
  return @logo if defined?(@logo)
  @logo = begin
    ...
  end
end
```

with the reject path falling through to `nil` instead of `return nil`. Worth giving `logo_size` the same treatment while you are in there, since it has the identical bug and is what multiplies this by five.

app/uploaders/image_uploader.rb
```diff
@@ -8,6 +8,41 @@ class ImageUploader < Shrine
...
+  # CVE; what the list buys is rejecting content that is not an image at all,
+  # with a clean validation error. HEIC/HEIF/AVIF and WebP matter because avatars
+  # come off phones, GIF and TIFF because they were uploadable before this list
+  # existed and turning a working upload into an error is the worse failure.
+  #
+  # The MIME type comes from marcel content analysis (determine_mime_type
+  # plugin), not from the client-supplied header — which matters for
+  # User#image_b64= and Actor#image_b64=, where Base64StringIO carries the
+  # content type the client declared in the data URI.
+  MIME_TYPES = %w[
```
SCjona, line 38: Small gap in the phone-camera coverage this list is justified by. marcel keys HEIF off the `ftyp` brand and emits five distinct types, not three (`marcel/tables.rb`):

```ruby
['image/avif',          [[4, 'ftypavif']]],
['image/heic',          [[4, 'ftypheic'], [4, 'ftypheix']]],
['image/heic-sequence', [[4, 'ftyphevc'], [4, 'ftyphevx']]],
['image/heif',          [[4, 'ftypmif1']]],
['image/heif-sequence', [[4, 'ftypmsf1']]],
```

`image/heic-sequence` and `image/heif-sequence` are not on the allowlist. Those are the `hevc`/`hevx`/`msf1` brands — the multi-image HEIF containers that burst and Live-Photo-adjacent exports land in. libvips reads them through the same `heifload`, which is not in the untrusted set, so they are a 422 on a file that would otherwise have processed fine. The plain-still case (`ftypheic`) is covered, which is the common one, so this is minor — but two more entries would make the "avatars come off phones" argument hold all the way.

Separately, and this is why the specs would not have surfaced it: both new suites gate the HEIC/WebP examples on the *saver* being compiled in (`image_buffer` → `skip`, and `next` in the blocking spec). On a libvips without a HEIF encoder those examples vanish, and nothing asserts that `image/heic` is the string marcel actually returns — the allowlist key could be misspelled and CI would stay green. A detection-only assertion against a fixed `ftyp` header needs no encoder:

```ruby
it 'uses the MIME types marcel actually emits for HEIF' do
  expect(Marcel::MimeType.for(StringIO.new("\x00\x00\x00\x18ftypheic\x00\x00\x00\x00"))).to eq('image/heic')
end
```

SCjona (commented):
Re-reviewed at `1a51936`. Both blocking items from the last round are properly closed, and I verified the mechanisms rather than taking the descriptions on faith:

- **Mailer gate.** `Marcel::MimeType.for(_tmp)` runs before `Vips::Image.new_from_file`, on a rewound Tempfile, with no `name:`/`declared_type:` — so it is pure magic-byte detection and the data-URI content type genuinely cannot influence it. The reject path closes and unlinks the tempfile, `attachments.inline` is never touched, and `nil` is what the layout expects (`app/views/layouts/mailer.html.erb:129` guards on `<% if @logo %>`). Worth recording that this also fixes a *hard* failure, not just the librsvg exposure: pre-PR an SVG/BMP/ICO logo raised an uncaught `Vips::Error` inside `prepare_logo`, which took the whole mail down.
- **`remove_invalid`.** Traced in shrine 3.8.0: `deassign` calls `destroy` → `file&.delete` (`attacher.rb:210`) plus `delete_derivatives`, from an `ensure`, so `errors` survives and the 422 via shrine-mongoid is unaffected. And the new `:store` assertions go through `record.image =`, which is the path that was hiding this.
- **`Vips.block('VipsForeignLoadSvg', false)`.** The parent-class form is correct — `vips_operation_block_set` maps over the named type and everything below it, and the abstract `VipsForeignLoadSvg` is a registered GType, so file/buffer/source are all covered. Turning the silent-no-op caveat into a comment that points at the spec is the right resolution.
- **Shared `MIME_TYPES`.** Lazy constant lookup inside `Attacher.validate` is fine under Zeitwerk; the drift concern is gone.

Nothing blocking left. Four things below, in descending order of how much I would want them dealt with. Only the first is one I would resolve *before* merging, and that is because it concerns data that already exists in production rather than the code.

Two minor notes that do not warrant an inline:

- `require 'marcel'` (`devise_mailer.rb:1`) works, but marcel is not a declared dependency — it reaches the lockfile only via `activestorage` (which `config/application.rb` deliberately does not load) and `caxlsx`. Bundler activates it either way, so this is not a bug today; `gem 'marcel'` in the Gemfile would make the new direct dependency explicit and survive either of those going away.
- The `loadable?` positive-control point from last round is still open in `spec/lib/vips_blocking_spec.rb`, and still non-blocking. Flagging only so it does not get lost, since it is the example carrying the central security assertion. The last inline comment below is a related instance of the same pattern.

spec/lib/vips_blocking_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    # ppmload — passing for the wrong reason on the one thing that matters here.
+    begin
+      Vips.block('VipsForeignLoadPpm', false)
+      expect(loadable?(file)).to be(true),
+                                 'ppmload is missing from this libvips build, so the block assertion proves nothing'
+    ensure
+      Vips.block('VipsForeignLoadPpm', true)
+    end
+
+    expect(loadable?(file)).to be false
```
SCjona, line 49: The control itself works, but it runs *before* the assertion it protects — and its `ensure` sets the flag the assertion then measures. So this example no longer tests boot state at all.

Walk it with the initializer's blocking removed entirely (both `Vips.block_untrusted(true)` and the `require` that also calls it):

1. `Vips.block('VipsForeignLoadPpm', false)` — no-op, ppm was already unblocked.
2. `expect(loadable?(file)).to be(true)` — passes.
3. `ensure` → `Vips.block('VipsForeignLoadPpm', true)` — blocks ppm *by hand*.
4. `expect(loadable?(file)).to be false` — passes, because of step 3.

Green, with every untrusted loader open in production. What the example now proves is "ppmload exists" and "`Vips.block(name, true)` blocks", neither of which is the property this file exists to guard. Before `f2804a6` line 49 measured the initializer; now it measures the `ensure`.

To be clear about where this came from: the snippet I gave you last round had exactly this ordering, so this is my error carried into the branch, not a regression you introduced. The fix is to read boot state before anything touches the flag:

```ruby
it 'blocks untrusted loaders' do
  # Netpbm stands in for the whole untrusted set (BMP, ICO, PSD, JPEG XL,
  # JPEG 2000, FITS, Matlab, OpenSlide, ImageMagick delegates).
  file = tempfile("P3\n1 1\n255\n0 0 0\n", '.ppm')

  # Boot state first. Once the control below has flipped the flag by hand, any
  # later assertion measures Vips.block rather than the initializer.
  expect(loadable?(file)).to be false

  # Positive control: libvips reports a blocked loader and a loader that was
  # never compiled in through the same "not a known file format" error, so
  # without this the assertion above would be green on a build without
  # ppmload — passing for the wrong reason on the one thing that matters here.
  begin
    Vips.block('VipsForeignLoadPpm', false)
    expect(loadable?(file)).to be(true),
                               'ppmload is missing from this libvips build, so the block assertion proves nothing'
  ensure
    Vips.block('VipsForeignLoadPpm', true)
  end
end
```

Same two assertions, same control, but now each one can only be satisfied by the thing it names. Worth checking it fails for the right reason before you trust it: comment out `Vips.block_untrusted(true)` in the initializer and the first assertion should go red while the control stays green.

Carrying this correction to samedis-care-backend#4108 too, since you said the control is going in there next round.

SCjona (commented):
Re-reviewed at `f2804a6`. The production code is unchanged from `1a51936`, which I already traced and signed off on — the mailer gate, `remove_invalid`, the shared `MIME_TYPES` and the parent-class SVG unblock all still hold. This round is only about the two spec changes.

**The camera-format `skip` is right.** `filter_map` + `next` drops the extension, and `skip` after the assertions marks the example pending rather than green, so a build without either saver is now visible in the run output instead of indistinguishable from a real check.

**The positive control has teeth, but it is sequenced so that it removes the teeth from the assertion it was added to protect.** That is the one thing below, and the ordering flaw is in the snippet I handed you last round — inline on `spec/lib/vips_blocking_spec.rb`. Test-only, production behaviour is correct, so not blocking.

Also re-verified while I was in there, since `f2804a6` touches the file that carries the security assertion:

- `Vips.block('VipsForeignLoadPpm', true)` in the `ensure` restores the flag unconditionally, so the control leaves no state leak into the other examples — that part is fine.
- `DownloadUploader` (`app/uploaders/download_uploader.rb`) is the only other Shrine uploader and has no `derivatives` block and no libvips path, so it correctly needs no allowlist. `ImageUploader::Attachment` / `ActorImageUploader::Attachment` on `User#image` / `Actor#image` are the only two attachments that reach libvips, and both are covered. Tenants are `Actor`s, so `Api::V1::User::TenantsController`'s `image_b64` goes through `ActorImageUploader` and is covered too.
- The `at_least_libvips?(8, 13)` boot guard cannot bite in production: `ruby:4.0.1-trixie` + `libvips-dev` gives 8.15. Worth noting the guard also runs during `docker build`, in the `rails openapi_spec:api[v1]` step — so a base-image downgrade would fail the build rather than a deploy, which is the better place for it.

### Still open from the previous round

None of these are new; listing them with locations rather than repeating the inlines, since your last comment only spoke to the two spec items and I do not want them to drop.

1. **`LOGO_MIME_TYPES` is narrower than the status quo on the mailer path** (`app/mailers/devise_mailer.rb:12`). GIF is the case that matters — `gifload` is not untrusted, mail clients render it, and for any tenant that has one stored today the logo silently disappears from every mail with only a `logger.warn`. This is the one item I would still resolve before merging, and only because it concerns data that already exists in production: a console pass over the stored `config.mailer.logo_b64` prefixes across apps settles it either way. `validates_length_of :logo_b64, maximum: 1.megabyte` (`app/models/actors/app.rb:43`) is the only thing that has ever constrained that field, so the serializer documentation is not evidence that only PNGs are stored.
2. **`return nil` inside `@logo ||=`** (`app/mailers/devise_mailer.rb:227`) skips the memoization, and via `logo_size`/`logo_width`/`logo_height`/`logo_style` that is 5 × decode + Tempfile + `Marcel::MimeType.for` + 5 identical warnings per mail for a rejected logo. `defined?(@logo)` fixes it.
3. **`image/heic-sequence` and `image/heif-sequence`** are missing from `ImageUploader::MIME_TYPES` (`app/uploaders/image_uploader.rb:31`) — the `hevc`/`hevx`/`msf1` `ftyp` brands, same `heifload`. And no spec asserts the marcel strings themselves, because both suites gate the HEIF examples on the *saver*; a detection-only assertion against a fixed `ftyp` header needs no encoder.
4. **`gem 'marcel'`** is not declared in the Gemfile, though `app/mailers/devise_mailer.rb:1` now requires it directly. It reaches the lockfile only via `activestorage` (deliberately not loaded) and `caxlsx`. Not a bug today.

spec/lib/vips_blocking_spec.rb
SCjona, line 49: Still open from the previous round — I think this one was read as "is the positive control in?" rather than "where does it sit?". The control is in and it has teeth; the problem is that it runs *before* the assertion it protects, and its `ensure` sets the very flag line 49 then measures.

Walk it with the initializer's blocking removed entirely (both `Vips.block_untrusted(true)` and the `require` that also calls it):

1. line 42 `Vips.block('VipsForeignLoadPpm', false)` — no-op, ppm was already unblocked.
2. line 43 — passes.
3. line 46 `ensure` → `Vips.block('VipsForeignLoadPpm', true)` — blocks ppm **by hand**.
4. line 49 — passes, because of step 3.

Green, with every untrusted loader open in production. Line 49 is satisfied by line 46 regardless of boot state, so what this example proves is "ppmload exists" and "`Vips.block(name, true)` blocks" — neither of which is the property this file exists to guard. The header comment at lines 5-10 is accurate, but it only covers the *SVG* examples: those do measure boot state, because nothing in the file unblocks svgload by hand. The untrusted-blocking assertion is the one that no longer does.

Fix is to read boot state first, before anything touches the flag:

```ruby
it 'blocks untrusted loaders' do
  # Netpbm stands in for the whole untrusted set (BMP, ICO, PSD, JPEG XL,
  # JPEG 2000, FITS, Matlab, OpenSlide, ImageMagick delegates).
  file = tempfile("P3\n1 1\n255\n0 0 0\n", '.ppm')

  # Boot state first. Once the control below has flipped the flag by hand, any
  # later assertion measures Vips.block rather than the initializer.
  expect(loadable?(file)).to be false

  # Positive control: libvips reports a blocked loader and a loader that was
  # never compiled in through the same "not a known file format" error, so
  # without this the assertion above would be green on a build without
  # ppmload — passing for the wrong reason on the one thing that matters here.
  begin
    Vips.block('VipsForeignLoadPpm', false)
    expect(loadable?(file)).to be(true),
                               'ppmload is missing from this libvips build, so the block assertion proves nothing'
  ensure
    Vips.block('VipsForeignLoadPpm', true)
  end
end
```

Same two assertions, same control, but each can only be satisfied by the thing it names. Worth confirming it fails for the right reason before trusting it: comment out `Vips.block_untrusted(true)` in the initializer and the *first* assertion should go red while the control stays green. Test-only, production behaviour is correct, so not blocking — but this is the example carrying the central security claim of the PR.

app/mailers/devise_mailer.rb
```diff
@@ -242,6 +253,36 @@ def logo_style
     end
   end
 
+  def build_logo
+    _logo_b64 = app.config.try(:mailer).try(:logo_b64) rescue nil
+    return nil unless _logo_b64.present?
+
+    _file = Base64StringIO.from_base64(_logo_b64, 'logo')
+    _file.try :rewind
+    _file.binmode
```
SCjona, line 262: The MIME gate is right, and it does convert the SVG/BMP/ICO/non-image cases from "uncaught `Vips::Error` takes the whole mail down" into a `nil` the layout handles. Two failure modes inside `build_logo` still take the mail down, though, and since this method now exists precisely to keep bad logo bytes away from mail rendering they are worth closing here rather than in the deferred `rescue_from Vips::Error` follow-up.

**1. `_file` can be `nil` at this line.** `logo_b64` is a plain `String` field whose only constraint is `validates_length_of ... minimum: 1, maximum: 1.megabyte` (`app/models/actors/app.rb:43`), and `Actors::App.ensure_base64` returns a `String` argument completely untouched (`app/models/application_document.rb:260-263`) — no format check at all. So an app admin can `PATCH` `config.mailer.logo_b64` to any string, and `Api::V1::AppAdminController::PERMIT_UPDATE` permits exactly that.

With `logo_b64 = "hello"`: `Base64StringIO.from_base64` does `URI.parse("hello")` → `opaque` is `nil` → `_url.try(:opaque).blank?` is true → `return nil` (`lib/base_64_string_io.rb:11-13`). Line 261 `_file.try :rewind` swallows that, but line 262 `_file.binmode` raises `NoMethodError`, uncaught, inside `prepare_logo` — every mail that app sends fails, including password reset. A near-miss variant reaches it too: `"data:image/png,abc"` passes the URL check, then `split(';base64,', 2)` yields no data half and `Base64.decode64(nil)` raises (`base_64_string_io.rb:17,32`).

**2. `Vips::Image.new_from_file` at line 279 can still raise for content the gate accepts.** A truncated or corrupt PNG is the mundane case. The sharper one is that this round widened `LOGO_MIME_TYPES` to include `image/heic`, `image/heif`, `image/avif` and `image/tiff` — the HEIF loaders are the optional part of a libvips build, so a MIME type on the list that the deployed libvips cannot decode moves the failure from a `logger.warn` to a dead mailer. Your own specs treat those savers as optional for exactly this reason. On that path the Tempfile is also leaked: lines 281-282 never run, so the fd and the file stay until the finaliser.

Both are pre-existing in substance — the old `logo` had the same shape — so not blocking. But the body is new code with a stated robustness goal, `logo`/`logo_size` already treat `nil` as meaningful, and the whole thing is one wrapper:

```ruby
def build_logo
  ...
rescue StandardError => e
  Rails.logger.warn("DeviseMailer: unusable mail logo (#{e.class}: #{e.message})")
  nil
ensure
  if _tmp
    _tmp.close unless _tmp.closed?
    _tmp.delete
  end
end
```

with the two explicit `_tmp.close`/`_tmp.delete` pairs dropped in favour of the `ensure` (note `_tmp.read` on line 280 has to stay before it). That turns "an admin can break every mail for their app with a malformed field" into a log line, which is the same trade the MIME gate already makes.

spec/uploaders/image_uploaders_spec.rb
```diff
@@ -0,0 +1,203 @@
...
+          expect(Marcel::MimeType.for(io)).to eq(mime_type)
+          expect(ImageUploader::MIME_TYPES).to include(mime_type)
+        end
+      end
+
+    %w[ftypheix ftyphevc ftypmsf1].each do |brand|
+      it "resolves #{brand} to video/quicktime, so the allowlist rejects it" do
+        io = StringIO.new("\x00\x00\x00\x18#{brand}\x00\x00\x00\x00".b)
+        detected = Marcel::MimeType.for(io)
+        expect(detected).to eq('video/quicktime')
```
SCjona, line 107: Minor, and only about the three negative examples — the three positive ones above are the valuable half, because they pin that the allowlist keys are the strings marcel actually emits without needing an encoder.

`expect(detected).to eq('video/quicktime')` pins a marcel *ordering quirk* as expected behaviour: the brand-specific HEIF entries exist in `tables.rb`, they are simply shadowed by the generic `[4, 'ftyp']` quicktime pattern that sits ahead of them. If a future marcel reorders that table — which would be a straight bug fix on their side — `ftypheix` starts resolving to `image/heic`, and these three go red on a routine dependency bump with a message that reads as "the allowlist is broken" when in fact the situation improved and the correct response is to revisit whether the sequence brands belong on the list.

The second assertion in each example (`expect(ImageUploader::MIME_TYPES).not_to include(detected)`) is the one that states the property you care about and holds either way. Either keep just that one, or leave the `eq` in with a custom message so the next person reads the right conclusion:

```ruby
expect(detected).to eq('video/quicktime'),
                    "marcel now resolves #{brand} to #{detected} rather than shadowing it with the generic " \
                    'ftyp/quicktime pattern — re-evaluate whether the HEIF sequence types belong on the allowlist'
```

The comment at lines 86-92 already carries the full explanation, so this is just about making the failure output point at it.

SCjona (commented):
Re-reviewed at `ba834d6`. All five items from last round are genuinely closed, and I checked the mechanisms rather than the descriptions:

- **`LOGO_MIME_TYPES` widened** (`devise_mailer.rb:23`). `ImageUploader::MIME_TYPES - ['image/svg+xml']` is the right shape — expressed as a subtraction, so the two lists cannot drift, and GIF is pinned by its own example. The `attachments.inline['logo.png']` mislabelling is recorded as a pre-existing quirk, which is the correct scope call.
- **Memoisation** (`devise_mailer.rb:223-235`). `defined?(@logo)` / `defined?(@logo_size)` both memoise `nil` and `''`, so `prepare_logo → logo_size → logo_width/height/style` collapses to one `build_logo`. Extracting the body was the right way to keep the guard clauses flat — the old `return nil` inside `@logo ||=` could not have been fixed in place. `logo_width` still returns `nil` from `''.split('x')[0]` and so never memoises, but with `logo_size` cached that is now free.
- **HEIF brands.** Your measurement is right and I reproduced the reasoning against marcel 1.2.1: `Marcel::Magic.by_magic` returns on the first table hit, and `video/quicktime`'s generic `[4, 'ftyp']` entry precedes the brand-specific ones, so `heix`/`hevc`/`msf1` never surface as an image type. Adding `image/heic-sequence` would indeed have been dead weight. The six encoder-independent examples are the real fix here — they assert the allowlist keys are the strings marcel emits, which the saver-gated examples could not.
- **`remove_invalid` write-through** documented at `image_uploader.rb:15-21`, including that `User#tenant_access_group_ids` is the reachable trap.
- **`gem 'marcel'`** in the Gemfile plus `DEPENDENCIES`; `marcel (1.2.1)` was already in the `GEM` section via `activestorage`/`caxlsx`, so the lockfile needed nothing else.

Also re-verified that the process-wide unblock is still fully covered: `Vips` appears only in the two uploaders' `ImageProcessing::Vips` pipelines and `devise_mailer.rb:279`, and the uploader validations run on `attach`, while derivatives are built later from `before_save :image_derivatives` (`user.rb:248`) and the Actor callback (`actor.rb:357`) — so the allowlist is genuinely in front of libvips on both upload paths.

Three things below, none blocking. The first is the item from the previous round that I do not think landed — your update comment reads it as "does the control exist", but the point was about where it sits.

One forward-looking note not worth an inline: `LOGO_MIME_TYPES` encodes "the uploader list minus SVG", not "the uploader list minus whatever the initializer unblocks". Today those are the same set and `devise_mailer_logo_spec.rb` pins the SVG half. If a second untrusted loader is ever unblocked in `config/initializers/vips.rb`, it becomes mailer-reachable silently. A one-line spec asserting `LOGO_MIME_TYPES` excludes every format the initializer unblocks would keep the invariant rather than the instance — follow-up territory.

Still open and unchanged as agreed follow-ups: no `validate_max_size` on either uploader (decompression bombs through the trusted PNG loader, or an SVG with huge `width`/`height`), and no `rescue_from Vips::Error`.

SCjona (approved):

PR #275, opened 2026-08-12
Clean up orphaned group mappings when clearing a user's tenant access

neongrau:
## What

`Api::V1::AccessControl::Tenant::UsersController#destroy` removes a user's access to a tenant by setting `access_group_ids = []`, which is supposed to unmap the user from every access group. `set_access_group_ids=` only iterates `Actors::Group.where(parent_ids: tenant.id)` — groups that currently exist — so it can never reach an `Actors::Mapping` whose target group was deleted/merged elsewhere without cascading the cleanup. The endpoint reports `success: true` while the dead group id keeps getting echoed back on every subsequent read, because `determine_tenant_access_group_ids` rebuilds the cache straight from the (still-dangling) mapping.

Confirmed live against the production report in Samedis-care/samedis-care-issues#2394: both access groups a specific user was cached as belonging to no longer exist as `Actors::Group` documents at all, yet their `Actors::Mapping` join records were still present.

Adds `User#unmap_orphaned_group_memberships!`, called from `set_access_group_ids=`, which removes any mapping for the actor whose parent group no longer resolves at all — regardless of whether the caller explicitly deselected it.

## Why `.delete` and not `.destroy`

`.destroy` fires `Actors::Mapping`'s `after_destroy :user_cache_expire!`, which calls back into the *same* user's cache/save machinery while still inside that user's own `before_save` — a reentrant `SystemStackError`, caught by the regression spec before this ever reached review. `.delete` skips that callback, matching the existing pattern used by `unmap_from!` a few lines above and by the sibling `Actors::Mapping.cleanup_orphans!`. Nothing is lost: `do_before_save` already calls `record.cache_expire!` right after this method returns.

## Verified

- New spec (`spec/models/user_set_access_group_ids_spec.rb`) reproduces the exact bug: maps a user into a group, deletes the group directly (the state the underlying data-integrity gap leaves behind), then confirms `access_group_ids = []` actually clears it.
- Confirmed the spec is a real regression test, not decoration — reverted just the fix and reran:
  ```
  1) ... removes the dangling mapping when access is fully cleared
     expected `Mongoid::Criteria#count` to have changed from 1 to 0, but did not change
  2) ... no longer reports the dead group id afterwards
     expected `["...").empty?` to be truthy, got false
  3 examples, 2 failures
  ```
  Restored the fix, both green again.
- A third case confirms selecting a still-existing group is unaffected (0→1 mapping created normally, not swept up as orphaned).
- `rubocop app/models/user.rb` — zero new offenses in the touched lines (the file has 163 pre-existing offenses elsewhere, untouched by this diff).
- Ran the two other specs that touch this same model area (`spec/migrations/repair_missing_standard_user_mappings_spec.rb`, `spec/lib/apac_migration/leak_guard_spec.rb`) — no new failures; confirmed the 4 failures in the migration spec are pre-existing by running the same file against unmodified `origin/main`.
- Security review: the new method is only reachable through the already-authorized, already-tenant-scoped `destroy`/`update` actions (`cando: ~/access-control.writer`, `tenant_id` intersected against the caller's own tenant memberships). It cannot cross tenant boundaries to delete a *live* mapping under another tenant — a still-existing group anywhere always fails the `exists?` check and is skipped; only mappings whose target is already unrecoverably gone are touched.

## Deliberately left out

- The one-off data cleanup for other tenants/users already carrying this kind of dangling reference (this PR only stops it from recurring going forward) — tracked as a separate task, not a code change.
- `determine_tenant_access_group_ids`'s `.dig(:tenant_access_group_ids)` (symbol key on a hash whose keys likely come back as strings from the raw aggregation) — flagged during investigation as a possibly-separate latent issue, unverified, and would manifest as the *opposite* symptom (access wrongly appearing empty) if real. Not touched here to keep this diff to the one confirmed bug.
- N+1 query in the new method's `.each` loop (one `exists?` check per mapping) — consistent with this method's existing style (`map_into!`/`unmap_from!` per group, unbatched), not a new performance profile. Worth revisiting if a user with very large mapping counts turns up.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -903,7 +903,7 @@ class User < ApplicationDocument
     @access_group_ids = tenant_access_group_ids[tenant_context.to_s] = _resulting_ids
   end
 
-  # Mappings whose target group no longer exists at all (e.g. the group was
+  # Mappings whose target actor no longer exists at all (e.g. the group was
   # deleted/merged elsewhere without cascading the cleanup). These can never
   # show up in tenant_groups_available above, so the unmap loop above can
   # never reach them via unmap_from! — without this, a full removal
@@ -911,9 +911,15 @@ class User < ApplicationDocument
   # Api::V1::AccessControl::Tenant::UsersController#destroy) can never
   # actually clear them, and the same dead group id gets echoed back forever
   # on every subsequent read. Scoped to the whole actor rather than just the
-  # current tenant: once the target group is gone there is no reliable way
-  # to recover which tenant it belonged to, and a mapping to a nonexistent
-  # group is garbage regardless of tenant.
+  # current tenant: once the target is gone there is no reliable way to
+  # recover which tenant it belonged to, and a mapping to a nonexistent
+  # actor is garbage regardless of tenant.
+  # Existence check is against `Actor`, NOT `Actors::Group` — mappings are
+  # not restricted to groups (MAPPABLE_TYPES includes ou/position/tenant
+  # too, e.g. via Api::V1::Apps::Users::ActorsController#create mapping a
+  # user into an arbitrary actor). Actors::Group.where(...) is STI-scoped
+  # to _type: "Actors::Group", so a live Ou/Organization/Tenant/Position
+  # parent would wrongly read as nonexistent and get deleted here.
   # Uses `.delete` (matching unmap_from! and Actors::Mapping.cleanup_orphans!
   # above) rather than `.destroy` — `.destroy` fires Actors::Mapping's
   # `after_destroy :user_cache_expire!`, which calls back into this same
@@ -923,7 +929,7 @@ class User < ApplicationDocument
   # record.cache_expire! right after this method returns.
   def unmap_orphaned_group_memberships!
     Actors::Mapping.where(map_actor: actor).each do |mapping|
-      next if Actors::Group.where(_id: mapping.parent_id).exists?
+      next if Actor.where(_id: mapping.parent_id).exists?
 
       mapping.delete
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -899,7 +899,34 @@ class User < ApplicationDocument
         group.unmap_from!(actor, cache_expire: false)
       end
     end
+    unmap_orphaned_group_memberships!
     @access_group_ids = tenant_access_group_ids[tenant_context.to_s] = _resulting_ids
+  end
+
+  # Mappings whose target group no longer exists at all (e.g. the group was
+  # deleted/merged elsewhere without cascading the cleanup). These can never
+  # show up in tenant_groups_available above, so the unmap loop above can
+  # never reach them via unmap_from! — without this, a full removal
+  # (`access_group_ids = []`, see
+  # Api::V1::AccessControl::Tenant::UsersController#destroy) can never
+  # actually clear them, and the same dead group id gets echoed back forever
+  # on every subsequent read. Scoped to the whole actor rather than just the
+  # current tenant: once the target group is gone there is no reliable way
+  # to recover which tenant it belonged to, and a mapping to a nonexistent
+  # group is garbage regardless of tenant.
+  # Uses `.delete` (matching unmap_from! and Actors::Mapping.cleanup_orphans!
+  # above) rather than `.destroy` — `.destroy` fires Actors::Mapping's
+  # `after_destroy :user_cache_expire!`, which calls back into this same
+  # user's cache/save machinery while we're still inside this user's own
+  # before_save, causing infinite reentrant recursion. `.delete` skips that
+  # callback entirely, which is fine here since do_before_save already calls
+  # record.cache_expire! right after this method returns.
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -899,9 +899,36 @@ def set_access_group_ids=(tenant_groups_selected)
...
+  # Uses `.delete` (matching unmap_from! and Actors::Mapping.cleanup_orphans!
+  # above) rather than `.destroy` — `.destroy` fires Actors::Mapping's
+  # `after_destroy :user_cache_expire!`, which calls back into this same
+  # user's cache/save machinery while we're still inside this user's own
+  # before_save, causing infinite reentrant recursion. `.delete` skips that
+  # callback entirely, which is fine here since do_before_save already calls
+  # record.cache_expire! right after this method returns.
+  def unmap_orphaned_group_memberships!
+    Actors::Mapping.where(map_actor: actor).each do |mapping|
+      next if Actors::Group.where(_id: mapping.parent_id).exists?
```
SCjona: **Blocking — data loss:** the existence check is scoped to `Actors::Group`, but a user actor's mappings are not restricted to groups.

`Actors::Group` is an STI subclass of `Actor`, so Mongoid adds `_type: {$in: ["Actors::Group"]}` to this criteria. Any mapping whose parent is a *live* `Actors::Ou`, `Actors::Organization`, `Actors::Tenant` or `Actors::Enterprise` therefore returns `exists? == false` and gets `.delete`d as if it were orphaned.

Those mappings are a real, supported concept, not a hypothetical:
- `Api::V1::Apps::Users::ActorsController#create` (app/controllers/api/v1/apps/users/actors_controller.rb:18-19, routed at config/routes/v1.rb:53) resolves the target as `Actor.where(parent_ids: current_app_id).find(actor_id)` — **any** actor type — and calls `actor.map_into!(target_user.actor)`. Same in `Api::V1::Apps::Tenants::Users::ActorsController#create` (config/routes/v1.rb:74), whose own security note explicitly talks about "mapping users into arbitrary actors".
- `Actor::MAPPABLE_TYPES = %i(user group ou position tenant)` (app/models/actor.rb:43) and the `groups_and_ous` scope (app/models/actor.rb:471) confirm non-group parents are expected.

**Failure scenario:** an admin maps user U into an `Actors::Ou` under app A via `POST .../users/:id/actors`. Later, any admin edits U's access groups in *any* tenant (or clears them via `Api::V1::AccessControl::Tenant::UsersController#destroy`). `do_before_save` (app/models/user.rb:289-290) → `set_access_group_ids=` → `unmap_orphaned_group_memberships!` walks **all** of U's mappings, finds no `Actors::Group` with the Ou's `_id`, and hard-deletes the mapping. Because the sweep is deliberately actor-wide rather than tenant-scoped, this also wipes non-group mappings belonging to *other* tenants and to the app level. `.delete` bypasses callbacks, nothing is logged, and there is no record of which actors were lost.

**Suggested fix:** check for the parent actor without constraining the type:

```ruby
next if Actor.where(_id: mapping.parent_id).exists?
```

That still satisfies the PR's goal — the #2394 mappings point at group ids that no longer exist as documents at all, so they fail an `Actor` existence check just the same. If the cleanup really should stay group-only, narrow the *scan* side instead (e.g. only consider mappings whose recorded parent was a group), but the existence probe itself must not be type-restricted.

SCjona (dismissed):
1 blocking finding(s):
- `unmap_orphaned_group_memberships!` treats every mapping whose parent is not an `Actors::Group` as orphaned and permanently deletes it — live mappings into an Ou / Organization / Tenant are silently destroyed on any access-group save.

SCjona (commented):
LGTM — no blocking findings.

Re-review of 391dda1. The single blocking finding from my previous review (https://github.com/Samedis-care/identity-management-backend/pull/275#discussion_r3765982496 — STI-scoped `Actors::Group` existence check hard-deleting live Ou/Organization/Tenant/Position mappings) is resolved at `app/models/user.rb:932` via the unscoped `Actor.where(_id: mapping.parent_id).exists?`.

Verified in the current code rather than just in the diff:
- `Actor` declares no `default_scope`, so the probe matches every STI subtype and also soft-deleted (`deleted: true`) parents — live non-group mappings are now skipped, and the new behaviour is strictly more conservative than the version I flagged, so it cannot introduce a new deletion path.
- The scan side `Actors::Mapping.where(map_actor: actor)` remains STI-scoped to mappings only, which is what you want here.
- `return unless tenant_groups_selected.is_a?(Array)` sits above the new call, so a non-array assignment cannot trigger a sweep.
- The `.delete` choice checks out: `Actors::Mapping` has `after_destroy :user_cache_expire!` → `map_actor.user.cache_expire!`, which would re-enter this user's save machinery from inside its own `before_save`; `unmap_from!` (app/models/actor.rb:809) and `cleanup_orphans!` use `.delete` for the same reason, and `do_before_save` expires the cache right after.
- The #2394 scenario is still covered — hard-deleted group ids resolve to no `Actor` document at all.

No new findings in the code that changed since the last round.

SCjona (approved):

PR #276, opened 2026-08-12
Reject raw HTML tags at intake on Actors::Tenant name fields

neongrau:
## What this is

Follow-up to Samedis-care/samedis-care-issues#2417 (Cognisys pentest, Stored HTML Injection) and its samedis-care-backend PR #4126. That PR found `Tenant#handle` in samedis-care-backend is echoed back from **this repo's** `/api/v1/user/tenant` response completely unvalidated — this backend is a second, independent source of the same payload class, not just a passive relay.

## Verified before writing any code

`Actor#get_name`'s `to_slug` conversion does **not** neutralize literal HTML tags — only incidentally strips quotes/slashes:

```ruby
'foo"><s>x<h1>y</h1>'.to_slug
# => "foo><s>x<h1>y<-h1>"   -- tags survive intact
```

`full_name` gets **zero** transformation at all when explicitly set — samedis-care-backend's `Api::V4::User::TenantsController#update` sends `short_name`/`full_name` straight from its own `name`/`name2` params, and this repo's `Actor#ensure_required_fields` only defaults `full_name` from other fields when it's blank, never sanitizes it when it's present.

## What it does

Ports the identical `SafeHtmlValidator` from samedis-care-backend (same technique: real HTML5 parser via the already-resolved `Loofah` gem, checking for parsed Element nodes — not a regex, not a text-diff comparison; both were tried and rejected there with documented failure modes in that repo's PR history for #2417).

Applied to `Actors::Tenant#name`/`#short_name`/`#full_name` specifically, **not** the shared `Actor` base class — those fields are used by every actor type (Users, Groups, Apps, OUs), and validating them there needs its own, separate audit this PR doesn't attempt.

Also adds `cleanup:strip_html_injection`, matching the sibling task added in samedis-care-backend PR #4123 for the same retroactive-validation concern (a tenant already containing a tag would otherwise fail validation on its next unrelated save).

## Verified

- Exact payload rejected on `Actors::Tenant.new(name:/short_name:/full_name:)` — checked all three fields independently.
- Ordinary text (`"acme-gmbh"`) passes cleanly.
- Cleanup task run end-to-end against a synthetic bad record (raw `collection.insert_one`, bypassing Mongoid — same technique as samedis-care-backend's migration-spec convention), confirmed dry-run reports it, `DRY_RUN=false` neutralizes it, re-run confirms idempotency, then the test record was deleted. Dry run against this environment's real data: 0 existing violations.
- `bundle exec rspec` on both new spec files: 12 examples, 0 failures. Existing `tenants_controller_spec.rb` (the cando/cross-tenant-enumeration regression guard) still passes unchanged: 4 examples, 0 failures.
- `bundle exec rubocop`: zero offenses on every touched/added file.

## Not in this PR

Not validating `name`/`short_name`/`full_name` on the shared `Actor` base class (Users, Groups, Apps, OUs) — same blast-radius reasoning as samedis-care-backend PR #4123 declining to touch `Notification#content`'s ~18 call sites. Worth its own audit if wanted.


app/validators/safe_html_validator.rb
```diff
@@ -0,0 +1,47 @@
...
+  def self.strip_tags(value)
+    return value unless contains_html_tag?(value)
+
+    Loofah::HTML5::DocumentFragment.parse(value).text
+  end
+
+  def self.contains_html_tag?(value)
+    return false if value.blank?
+
+    Loofah::HTML5::DocumentFragment.parse(value).children.any?(&:element?)
```
SCjona, line 79: The check only fires when the HTML5 fragment parser materializes an `Element` node as a **direct child** of the fragment. Three payload classes that are live markup in the downstream sink produce no element child, so they pass validation and get stored:

**1. Attribute breakout with no tag at all**

```
NewTestFacility" autofocus onfocus=alert(1) x="
```

No `<` anywhere, so `children.any?(&:element?)` is `false`. The pentest payload began with `">` precisely because the SCB sink is inside an attribute value — drop the `<s>`/`<h1>` tail and the remainder still breaks out of the attribute, but this validator accepts it.

**2. Start tags that fragment parsing discards**

`<body onload=alert(1)>` (same for `<html …>`, `<head …>`, and table-scoped tags such as `<td onmouseover=…>`). In the HTML5 "in body" insertion mode a `body`/`html`/`head` start tag is ignored when the parser runs as part of the fragment-parsing algorithm, and table-scoped tags outside a table are dropped — the fragment ends up with text children only, so `contains_html_tag?` returns `false`. Echoed into a real document, `<body onload=…>` merges its attributes onto the page's own `body` element and fires.

**3. EOF-truncated tag**

```
NewTestFacility<img src=x onerror=alert(1)
```

No trailing `>`, so the tokenizer hits `eof-in-tag`: that is a parse error and the pending tag token is **discarded**, leaving only the text node — validation passes. When the stored value is echoed unescaped into a page, the browser keeps consuming the *following page markup* as attributes until the next `>` and builds a live ``. That is exactly the "malformed/unclosed tags that lenient renderers still interpret as markup" case the comment on line 13 claims the parser-based approach covers.

**Suggested fix** — keep the parser check for the friendly message, but gate on the raw characters too:

```ruby
UNSAFE_CHARS = /[<>]/ # plus ["'] given the SCB sink is an attribute value

def self.contains_html_tag?(value)
  return false if value.blank?
  return true if value.match?(UNSAFE_CHARS)

  Loofah::HTML5::DocumentFragment.parse(value).children.any?(&:element?)
end
```

A character-class rejection is not "sanitizing HTML with a regex" — there is no tag grammar to bypass, and for a tenant `name`/`short_name`/`full_name` the characters `<` `>` have no legitimate use.

One consistency note if you tighten it this way: `strip_tags` (used by `cleanup:strip_html_injection`) relies on Loofah's `#text`, which *escapes* those characters rather than removing them — `NewTestFacility"><s>sss<h1>TEST</h1>` becomes `NewTestFacility&quot;&gt;sssTEST`. That still passes a `[<>"']` check (so the rake task stays idempotent), but it writes entity text into the tenant name, which renders literally as `&quot;&gt;` wherever the consumer escapes on output. Dropping the characters instead of escaping them avoids that.

SCjona (dismissed):
1 blocking finding:

- `SafeHtmlValidator` only rejects values that the HTML5 fragment parser turns into a direct Element child, so several payloads that are live markup in the downstream SCB sink (attribute breakout without any tag, `<body onload=…>`/table-scoped tags that fragment parsing ignores, and EOF-truncated tags like `foo<img src=x onerror=alert(1)`) pass the check and get stored — the intake control the PR adds does not hold for the payload class the pentest used.

lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,67 @@
...
+      bundle exec rake cleanup:strip_html_injection DRY_RUN=false
+  DESC
+  task strip_html_injection: :environment do
+    dry_run = ENV.fetch('DRY_RUN', 'true') != 'false'
+
+    puts '=' * 60
+    puts "STRIP PRE-EXISTING HTML INJECTION#{dry_run ? ' — DRY RUN' : ''}"
+    puts '=' * 60
+
+    tag_regex = /[<>]/
```
SCjona: The pre-filter is now narrower than the validator it is meant to front-run. `SafeHtmlValidator::UNSAFE_CHARS` is `/[<>"]/`, but this query matches only `/[<>]/`, so any tenant whose sole offending character is a double quote is never loaded into the loop — `contains_html_tag?` never gets a chance to judge it.

**Failure scenario:** an existing tenant named `Klinikum "Am Kurpark" GmbH` (or the pentest's own tag-less breakout variant, `NewTestFacility" autofocus onfocus=alert(1) x="`). Running `rake cleanup:strip_html_injection` prints `Nothing to do — no pre-existing HTML injection found.` The record stays untouched, yet `validates :name, :short_name, :full_name, safe_html: true` now rejects it on every subsequent save: an admin editing only the tenant logo via `Api::V1::User::TenantsController#update` gets `full_name must not contain HTML tags` on a field they never touched, and `Actor#ensure_defaults!` (`app/models/actor.rb:942`, `_actor.save!`) raises `Mongoid::Errors::Validations` on that subtree. That lockout is exactly what this task's own docstring says it exists to prevent.

The task comment calls the regex "just a cheap first pass, not the final decision" — but for quote-only values it *is* the final decision, because the candidate never reaches `contains_html_tag?`.

**Fix** — derive the pre-filter from the same constant so the two cannot drift apart again:

```ruby
tag_regex = SafeHtmlValidator::UNSAFE_CHARS
```

Worth re-running the dry run against production data afterwards: the earlier "0 existing violations" result was measured with the narrower `[<>]` filter, so it would not have surfaced quote-only names.

SCjona (dismissed):
1 blocking finding:

- The `cleanup:strip_html_injection` MongoDB pre-filter still matches only `[<>]`, while `SafeHtmlValidator::UNSAFE_CHARS` was widened to `[<>"]` — records whose only offending character is a double quote are never loaded, so the task reports "Nothing to do" while those tenants become unsaveable.

Previous round: the three parser-only bypasses (no-tag attribute breakout, `<body onload=…>`/table-scoped tags, EOF-truncated tag) are all closed by the character-class check — verified against the current code. The apostrophe exclusion is accepted as reasoned, and the `strip_tags` two-branch split resolves the entity-escaping note.

SCjona (commented):
LGTM — no blocking findings.

Round-2 finding resolved: `lib/tasks/cleanup.rake` now derives the MongoDB pre-filter from `SafeHtmlValidator::UNSAFE_CHARS` instead of a hand-maintained `[<>]` copy, so the two can no longer drift apart. The pre-filter is now provably a superset of what `contains_html_tag?` rejects: the character-class branch is matched exactly, and the parser branch can only fire on a value containing `<` (no `Element` node can be produced without one). A tenant whose only offending character is a double quote — the `Klinikum "Am Kurpark" GmbH` case — is now loaded and neutralized rather than silently left unsaveable. Thanks for verifying the before/after against a synthetic record and re-checking idempotency.

Round-1 findings remain closed — the three parser-only bypasses (no-tag attribute breakout, `<body onload=…>`/table-scoped tags, EOF-truncated tag) are all caught by the character-class check, re-verified against the current validator.

SCjona (approved):

PR #277, opened 2026-08-13
Stop before_save from reviving a soft-killed MFA token on any later save

neongrau:
## What

Cognisys pentest, July 2026 (PlexTrac `4120693929`): a session token remained valid after logout for accounts with MFA/OTP enabled. Refs #2422 (does not close it -- the samedis-care-backend companion PR, [samedis-care-backend#4129](https://github.com/Samedis-care/samedis-care-backend/pull/4129), fixes a related, separate gap on the same code path; #2422 should stay open until both land).

## Root cause

`Api::V1::Doorkeeper::TokensController#revoke` kills a token immediately for the "soft" logout branch (`token_type_hint: 'access_token'`) via `_token.update_attributes(expires_in: -1)`. For an OTP-enabled user, `IdentityManagementExtension`'s `before_save` callback unconditionally reset `expires_in` back to the full 30-day lifetime on *every* save -- including that very same revoke -- so logout silently never took effect. Non-MFA accounts were unaffected because the `else` branch never touches `expires_in` at all, matching exactly the MFA-only asymmetry Cognisys observed.

## Fix, and a correction along the way

First attempt guarded only the *same* save that set `expires_in` (checking `expires_in_changed?`). A local review round found a second, real trigger before this ever went out for external review: the refresh_token grant flow (`Api::V1::App::Doorkeeper::TokensController#create`) finds a soft-revoked token by its still-live refresh_token (surviving on purpose, per the remembered-account fast path), rotates that field, and calls `save!` on it -- an unrelated save where `expires_in_changed?` is false, so that narrower guard would still silently revive the token the next time its refresh token gets used. Reproduced live against the dev DB before fixing it.

The corrected guard keys off the token's *current* `expires_in` value (nil-or-positive → reset to full lifetime; already non-positive → leave it dead), not what this particular save changed. Also handles a second-order edge the correction introduced: a bare token creation with no explicit `expires_in` must still default to a real lifetime, not `nil` (which Doorkeeper's `Expirable#expired?` treats as "never expires" -- more permissive than intended). Every real token-creation path in this app passes `expires_in` explicitly, so this was unreachable in production, but the new test suite's own bare `create!` would otherwise have hit it silently.

Also removes the dead duplicate `if/else` this replaced (both arms set the identical value) and the now-fully-dead `OTP_TIME_TO_AUTHENTICATE` constant it was the last reference to -- the "short window to enter OTP before the token expires" feature it implies was never actually implemented (both arms always used the full lifetime), as far back as this repo's history goes. Left as an out-of-scope observation, not implemented here -- unrelated to what #2422 asks for.

## What was verified (live, against the real dev DB -- not just reasoning)

- Soft-revoke alone: `expires_in` persists as `-1` (was `2592000` before any fix).
- Soft-revoke followed by an unrelated save (refresh_token rotation, the actual refresh-grant code path): `expires_in` still `-1` (was `2592000` again with the first, narrower guard -- reproduced and confirmed before correcting it).
- A bare create with no explicit `expires_in` for an OTP-enabled user: defaults to the full lifetime, not `nil` (reproduced the nil-default gap with the interim guard, confirmed the correction closes it).
- Unaffected: non-MFA soft revoke, `authenticate_otp`'s own expiry bump, and the "hard" revoke branch (`_token.revoke`, unrelated to `expires_in`).
- `bundle exec rspec spec/models/doorkeeper/access_token_identity_management_extension_spec.rb` -- 7 examples, 0 failures. This repo had no test coverage anywhere for the revoke path before this PR.
- `bundle exec rubocop` on both changed files -- clean (only pre-existing, unrelated offenses in untouched lines).
- A security-focused review pass (token/state manipulation, privilege escalation via `im_otp_required`/`otp_satisfied?`, info leak) plus an 8-angle code review (correctness, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions) -- both are what caught the refresh-token-revival gap and the nil-default edge case above before this ever left local review.

## What's deliberately left out

- The dead `OTP_TIME_TO_AUTHENTICATE` short-window-before-OTP-confirmation feature: never implemented, unrelated to this issue, not built here.
- `self.user.otp_enabled?` runs an ungated `User` fetch on every token save while OTP is required -- pre-existing (not introduced by this diff), possibly deliberate (so a user who just enabled MFA retroactively affects already-issued tokens). Left alone; flagging in case it's worth its own look.

### DIFF
--- a/app/models/concerns/identity_management_extension.rb
+++ b/app/models/concerns/identity_management_extension.rb
@@ -15,8 +15,6 @@ module IdentityManagementExtension
 
     # token will be revoked after this many wrong OTPs
     OTP_MAX_TRIES = 3
-    # after first login this many seconds to supply OTP before token expires
-    OTP_TIME_TO_AUTHENTICATE = 300
     # Allow some seconds of drift (accepts shortly expired code) to compensate server clock drift
     OTP_DRIFT = 60
 
@@ -38,11 +36,26 @@ module IdentityManagementExtension
       # was provided for the token which is checked during JsonApiController#authorize
       if !self.im_otp_required.eql?(false) && self.user.otp_enabled?
         self.im_otp_required = true
-         # limit time to enter an otp before the token expires
-        if self.im_otp_provided
+
+        # Api::V1::Doorkeeper::TokensController#revoke kills a token immediately for
+        # the "soft" logout branch (token_type_hint: 'access_token') via
+        # update_attributes(expires_in: -1). Once expires_in is non-positive the
+        # token is dead and must STAY dead -- checking only "was expires_in changed
+        # in THIS save" is not enough: Api::V1::App::Doorkeeper::TokensController#create
+        # (the refresh_token grant) finds this same token by its still-live
+        # refresh_token, rotates that field, and calls save! on it -- an unrelated
+        # save that doesn't touch expires_in at all, so expires_in_changed? is false
+        # and this branch would otherwise revive a soft-killed token straight back to
+        # the full 30-day lifetime the moment its surviving refresh token is next
+        # used, resurrecting exactly the bug this fix exists to close (#2422).
+        #
+        # `nil` is not "dead" -- every real token-creation path sets expires_in
+        # explicitly, but a bare create without it must still default to a real
+        # lifetime rather than silently persisting nil, which Doorkeeper's
+        # Expirable#expired? treats as "never expires" (strictly more permissive
+        # than intended, not a safe fallback).
+        if expires_in.nil? || expires_in.to_i.positive?
           self.expires_in = Doorkeeper.configuration.access_token_expires_in
-        else
-          self.expires_in = Doorkeeper.configuration.access_token_expires_in #OTP_TIME_TO_AUTHENTICATE
         end
       else
         self.im_otp_required = false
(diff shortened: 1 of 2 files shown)


SCjona (commented):
LGTM — no blocking findings.

Verified the guard actually closes #2422's MFA path and that it does not break any legitimate lifetime reset:

- `authenticate_user!` (devise-doorkeeper) gates on `accessible?` → `!expired?`, so a persisted `expires_in: -1` genuinely rejects the token. Keying the guard off the token's *current* `expires_in` rather than `expires_in_changed?` is the right call — `Api::V1::App::Doorkeeper::TokensController#create` really does `save!` the previous token after rotating only `refresh_token` (tokens_controller.rb:40-41), which the narrower guard would have missed.
- Every path that must hand a token a live lifetime assigns `expires_in` explicitly before the callback runs, so none of them get stuck dead: `app/controllers/api/v1/app/doorkeeper/tokens_controller.rb:56` (this also covers the new token Doorkeeper's refresh grant builds by copying `expires_in` from the soft-killed predecessor), `identity_management_extension.rb:101` (`authenticate_otp`), `social_auth_controller.rb:64`, `omniauth_callbacks_controller.rb:34`. `reuse_access_token` is disabled (doorkeeper.rb:92), and `find_or_create_for` would not reuse an expired token in any case.
- The `nil` branch is needed as described: `content_acceptance_controller.rb:61` and `omniauth_callbacks_controller.rb:48` both call `expires_in.seconds.from_now`, so persisting `nil` would be worse than merely permissive.
- `OTP_TIME_TO_AUTHENTICATE` has no remaining references anywhere in the repo, and both arms of the replaced `if/else` set the identical value — the removal drops no behavior.
- Not re-raising the refresh-token-survives-soft-logout gap: it is pre-existing, deliberate per the remembered-account fast path, and already scoped to the companion PR.

SCjona (approved):

PR #278, opened 2026-08-14
Return a generic response for account recovery regardless of email existence

neongrau:
## Summary
Reverts the response-splitting introduced in `72341e1` ("be more specific", May 2024): the account-recovery endpoint (`POST /api/v1/*app/recovery`) returned a distinct 404 for a non-registered email vs. 400/200 for a registered one. That split is the exact oracle flagged by a Cognisys pentest (PlexTrac finding 807300379) as an email-enumeration vulnerability.

Both "no such user" and "user exists but has no recovery email configured" now render the same 400 response, identical to pre-72341e1 behavior. The success path (200, recovery email sent) is unchanged.

## Changes
- `app/controllers/api/v1/app/recoveries_controller.rb` — merge the two failure branches into one, with a `SECURITY / PENTEST NOTE` comment to stop this being re-split later
- `config/locales/{en,de,nl,fr,ru}.yml` — remove the now-unused `errors.user.record_not_found_error` key (added by `72341e1`, only referenced from the removed branch)
- `spec/controllers/api/v1/app/recoveries_controller_spec.rb` — new regression spec (controller previously had zero coverage): asserts identical 400 for both failure cases, 200 for the success case, and that 404 is never returned

## Residual risk (out of scope for this fix)
- 200 vs. 400 still differs for accounts that *have* recovery_email configured — a narrower oracle than full email enumeration, matching the accepted pre-2024 behavior.
- Minor timing side-channel: the success path does token generation + mail enqueue, the failure path returns immediately. Not part of this pentest finding (which is about response content).

Fixes Samedis-care/samedis-care-issues#2419


SCjona (commented):
LGTM — no blocking findings.

Verified:
- `app/controllers/api/v1/app/recoveries_controller.rb:21` — both failure cases (no user / user without `recovery_email`) now render the identical 400 `record_error` with `errors.user.recovery_token.recovery_email_unset`; no status or message signal about email existence remains on the failure path.
- The removed `errors.user.record_not_found_error` key has no remaining references anywhere in `app/`, `lib/`, or `config/`. The similarly named `json_api.record_not_found_error` used by `app/controllers/concerns/base_controller_methods.rb:112` is a different key and is untouched.
- The `errors.user` subtree still has children in all five locales (en/de/fr/nl/ru), so no empty-mapping breakage from the removals.
- Route `resource :recovery, only: [:create]` under `scope path: '*app'` is unchanged; the success path (`app_context=`, `send_recovery_instructions`, 200) is untouched.

The residual 200-vs-400 difference for accounts that do have a recovery email is already documented as accepted, out-of-scope risk in the PR description — not re-raised here.

SCjona (approved):

PR #279, opened 2026-08-14
Enforce password strength server-side, mirroring the frontend's zxcvbn check

neongrau:
## What

A Cognisys pentest (Informational, CVSS 0) found that `POST /api/v1/*/register` accepted arbitrarily weak passwords (e.g. `12345678`) because only `identity-management-frontend`'s `CreateAccount.tsx` enforced strength (zxcvbn score ≥ 2) — the API had no complexity check at all, only a length range.

This adds the same zxcvbn-based check server-side to `User`, applied to **every** path that can set a password:
- Registration (`Api::V1::Devise::RegistrationsController#create`)
- Self-service password change (`Api::V1::App::UserController#update`)
- Reset-by-token (`Api::V1::Devise::PasswordsController`)
- Admin/tenant-admin "set another user's password" (`Api::V1::UsersController` and friends) — **no trusted-admin exemption**, per issue discussion.

Also raises the minimum password length from 6 to 8, matching the frontend's `minPwLen`.

## Why the design looks the way it does

The obvious approach — just add a `validates` to the model — isn't sufficient here, because `User#set_password=` (used by admin endpoints, OAuth account provisioning, and the unconfirmed-account re-registration flow) persists the encrypted password **immediately** via Mongoid's `set()`, which bypasses callbacks/validations entirely (see this repo's own Mongoid `self.set()` guidance). A plain model validation would only report an error *after* the weak password was already written to the database for those paths.

So the fix has two parts:
1. `set_password=` now skips its own eager `set()` persist when the password is too weak — the weak digest is never written.
2. The new `validate_password_strength` model validation is **not** gated by `unless: :set_password` like the existing length/confirmation checks — it runs for every path, including the ones that go through `set_password=`, so the caller still gets a proper, localized validation error via the normal `errors`/`save` flow.

The only bypass is `skip_password_strength_validation`, set exclusively by `User.global_admin`'s fixed internal bootstrap password (`"#" * 18`, not a real user credential, predates this check) — never a permitted controller param.

Verified empirically (see commit): the PoC password `12345678` scores 0 (rejected), realistic passwords like `MyDogRex99` score 3 (pass), and OAuth's `SecureRandom.uuid`-derived auto-generated passwords score 4 (unaffected).

## Testing

- New spec: `spec/models/user_password_strength_spec.rb` (18 examples) — covers the exact PoC password, other common weak passwords, realistic passwords, the raised length minimum, `set_password=`'s eager-persist-skip behavior, the `global_admin` bootstrap exemption, and confirms `skip_password_strength_validation` is never in any controller's `PERMIT_CREATE`/`PERMIT_UPDATE`.
- Full regression pass across `spec/models/user_set_access_group_ids_spec.rb`, `spec/models/invite_spec.rb`, `spec/models/doorkeeper/access_token_identity_management_extension_spec.rb`, `spec/controllers/api/v1/app/user_controller_spec.rb`, `spec/controllers/api/v1/devise/omniauth_callbacks_controller_spec.rb`, `spec/migrations/repair_missing_standard_user_mappings_spec.rb` — no new failures (pre-existing failures in the migration and omniauth specs confirmed identical on unmodified `main`).
- `rubocop` on all changed files: zero new offenses (diffed against baseline).

Fixes Samedis-care/samedis-care-issues#2425


app/models/user.rb
```diff
@@ -731,6 +771,20 @@ def disable_devise_notifications!
...
+    # strength) must be rechecked HERE, before that write, not only in the validations
+    # below. `password_strong_enough?` alone is NOT sufficient: it only measures zxcvbn
+    # score, so a too-short (or, via the Zxcvbn::PasswordTooLong rescue, too-long) password
+    # can score as "strong" while still failing the length validation that runs later —
+    # letting a weak-but-short password slip through this eager persist even though the
+    # subsequent `.save` correctly reports "too short" (issue #2425 review finding).
+    # `password`/`password_confirmation` are already assigned above, so the subsequent
+    # `valid?`/`save` call still runs both validations and reports the real error.
+    acceptable = pwd.present? && self.class.password_length.cover?(pwd.length) &&
+                 self.class.password_strong_enough?(pwd, user_inputs: password_strength_user_inputs)
```
SCjona: The inline gate and the validation can disagree about the same password, and when the gate is the more permissive of the two the weak digest is persisted anyway — the same eager-persist hazard you fixed for length, arriving via `user_inputs` instead.

`password_strength_user_inputs` is `[email, first_name, last_name].compact` read off the record at the moment `set_password=` runs. `validate_password_strength` (line 807) recomputes it after every attribute has been assigned. `ActionController::Parameters#permit` builds its output in **PERMIT-list order**, and in all three admin controllers `:set_password` sits at index 7 while `:first_name`/`:last_name` are at 12/13:

- `app/controllers/api/v1/users_controller.rb:19` vs `:23`/`:24`
- `app/controllers/api/v1/apps/users_controller.rb:21` vs `:25`/`:26`
- `app/controllers/api/v1/apps/tenants/users_controller.rb:19` vs `:23`/`:24`

So `record.attributes = params_update` calls `set_password=` while `first_name`/`last_name` still hold the *old* values loaded from the database.

Concrete failure — existing user `bob@example.com`, currently named Bob Smith. An admin PATCHes `{first_name: "Jane", last_name: "Doe", set_password: "JaneDoeJaneDoe"}`, i.e. one form submit that renames the user and sets their password:

1. `set_password=` runs with `user_inputs: ["bob@example.com", "Bob", "Smith"]` → `JaneDoeJaneDoe` scores >= 2 → `acceptable` is true → `set(encrypted_password: ...)` fires. The record is persisted (`record_update` → `model_update.find(id)`), so that is an immediate `update_one` bypassing validations.
2. `first_name`/`last_name` are then assigned to Jane/Doe.
3. `json_api_controller#update:122` calls `record.validate` → `validate_password_strength` now runs with `["bob@example.com", "Jane", "Doe"]` → score < 2 → `too_weak`. `check_for_errors` returns at line 123, so `record.save!` (line 124) is never reached.

Net result: the API answers "password is too weak" and the rename is discarded — but the weak password is already live in MongoDB and Bob's previous password no longer works. Your own spec asserts this exact delta exists for `JaneDoeJaneDoe` with `%w[Jane Doe]` (`spec/models/user_password_strength_spec.rb`, the "is biased against the user's own email/name" example), so the two evaluations really can land on opposite sides of the threshold.

Suggested fix: stop doing the durable write from the attribute writer, since it can never see the record's final state. Have `set_password=` only record the pending password — Devise's `password=` already assigns `encrypted_password` in memory, so an ordinary `save` persists it — and move the `set(encrypted_password: ...)` into an `after_validation`/`before_save` hook guarded by `@set_password`. At that point `password_strength_user_inputs` is final and the validation has already passed, so gate and validation cannot disagree. If the eager `set()` has to stay in the writer for the no-notification path, at minimum re-check with the final inputs immediately before the write.

Gemfile
```diff
@@ -63,6 +63,9 @@ gem 'omniauth-microsoft_graph'
 gem 'omniauth-rails_csrf_protection'
 gem 'recaptcha', require: 'recaptcha/rails'
 gem 'tzinfo-data'
+# server-side password strength check (mirrors identity-management-frontend's
+# react-password-strength-bar, which also wraps zxcvbn) — issue #2425
+gem 'zxcvbn-ruby', require: 'zxcvbn'
```
SCjona, line 68: Non-blocking: `Zxcvbn.test` lazily builds a process-wide shared `Tester` on first call, and `Zxcvbn::Data#initialize` reads ~800 KB of frequency lists plus `adjacency_graphs.json` and builds a `Trie` per dictionary — all inside `DEFAULT_TESTER_MUTEX`. So the first `POST /register` (or password change) after each boot pays the whole dictionary load, and any concurrent password request blocks on that mutex until it finishes. Worth warming it once at boot (e.g. a `Zxcvbn.test('warmup')` in an initializer) so the cost lands on deploy rather than on a user request. Also worth knowing for capacity planning that the tries stay resident per Puma worker.

SCjona (dismissed):
1 blocking finding(s):
- `set_password=` gates its eager `set(encrypted_password:)` persist on `password_strength_user_inputs` **as of assignment time**, while `validate_password_strength` recomputes them **after** all attributes are assigned — so an admin update that changes `first_name`/`last_name` in the same request as `set_password` can write a weak digest to MongoDB and then still return "too weak" to the caller.

The rest of the change verified out fine: the gem API matches (`Zxcvbn.test(password, user_inputs)`, `Score#score` 0–4, `Zxcvbn::PasswordTooLong < ArgumentError` raised above `TesterBuilder::DEFAULT_MAX_PASSWORD_LENGTH = 256`, so the rescue is real and the 256-char cap bounds worst-case scoring time on the unauthenticated `/register` path); the `too_weak` key resolves under `mongoid.errors.models.user.attributes.password` in all five locales; `skip_password_strength_validation` appears in no PERMIT list; `User.global_admin`'s `first_or_create` does assign it before `set_password` (Mongoid processes mass-assignment in hash order); registration, `update_with_password`, `reset_password`-by-token and the generic `json_api_controller#update` all surface the new error instead of swallowing it; and `from_omniauth`'s `SecureRandom.uuid[0, 18]` clears both the new 8-char floor and the score threshold.

SCjona (commented):
LGTM — no blocking findings.

Re-review of 7945295. The one blocking finding from my previous review (assignment-order race between the eager `set(encrypted_password:)` persist and `validate_password_strength`) is resolved: `set_password=` now only stages the pending password, and the persist moved to `before_save :persist_set_password`. Mongoid’s `prepare_update` returns `false` on `invalid?(:update)` *before* `run_callbacks(:save)` fires, so a record that fails the strength/length check never reaches `persist_set_password` at all — the weak digest can no longer be written ahead of validation, regardless of mass-assignment order. The `Xylonqvist99` + same-request-rename regression spec covers exactly that path.

Verified on the newly changed code:

- **No collateral dirty-tracking loss from the deferred `set`.** This was my main worry with moving an atomic op into `before_save` — in Mongoid 9.1 (`mongoid (9.1.0)`), `Persistable#prepare_atomic_operation` does *not* call `post_process_persist`/`move_changes`; `process_atomic_operations` calls `remove_change(access)` per operated-on field only. So only `encrypted_password` is cleaned, and other pending changes assigned in the same request (`first_name`/`last_name` from the admin `PERMIT_UPDATE`, the `confirmation_sent_at`/`confirmation_token` resets in `new_with_session`, the OAuth token fields in `from_omniauth`) are still in `changed_attributes` when `init_atomic_updates` runs inside the yield, and still get written. Per-field `remove_change` is also exactly what keeps `send_password_change_notification?` false, as the comment claims.
- **Dropping the inline `password_length.cover?` gate (from 65ca041) is safe.** It is no longer load-bearing now that nothing persists before validation, and the length floor still holds on the `set_password=` update path: the model’s own `length: { in: 8..1024 }` is `on: :create, unless: :set_password`, but `devise :validatable` contributes an unconditional `validates_length_of :password, within: Devise.password_length` (now `8..128`), which fires on update whenever `password` is present. Because Mongoid sets `i18n_scope :mongoid`, that error resolves to `mongoid.errors.models.user.attributes.password.too_short` — the same message the specs assert — and it also catches the >128-char case that `password_strong_enough?` deliberately rescues to `true`.
- **New-record path unaffected.** `persist_atomic_operations` is guarded by `persisted?`, so the `before_save` `set` is a DB no-op on create (`User.global_admin`, registration); `password=` already put the digest in `attributes`, and the insert writes the full document. `skip_password_strength_validation` is now order-independent anyway, since it is only read at validation time.
- Everything I verified in the previous round (gem API, `Zxcvbn::PasswordTooLong < ArgumentError` above the 256-char cap, the `too_weak` key in all five locales, no `skip_password_strength_validation` in any PERMIT list, error surfacing on register/`update_with_password`/reset-by-token/`json_api_controller#update`) is unchanged by this commit.

The `config/initializers/zxcvbn.rb` warmup addresses my previous non-blocking note as suggested.

SCjona (approved):

PR #280, opened 2026-08-14
Filter active/expired logins by the token's own expiry, not its age

neongrau:
## What

Cognisys pentest follow-up (July 2026, PlexTrac `4120693929`), **related gap 4** of the analysis on Samedis-care/samedis-care-issues#2422: a session that had been properly logged out still showed up as an *active* login. Refs #2422, does not close it -- gaps 1 and 2 of that analysis are untouched (see below).

The reported finding itself, and the samedis-care-side gap 3, are already merged: #277 (`1e8cb51`) and [samedis-care-backend#4129](https://github.com/Samedis-care/samedis-care-backend/pull/4129).

## Root cause

`User#active_logins` and `#expired_logins` derived a token's expiry window from `created_at` plus the **globally** configured `access_token_expires_in`, and never looked at the token's own `expires_in`.

`Api::V1::Doorkeeper::TokensController#revoke` kills a token for the "soft" logout branch (`token_type_hint: 'access_token'`) by writing `expires_in: -1` while deliberately leaving `revoked_at` nil -- the refresh token has to survive, because that is what powers the remembered-account fast path on the login page. So a just-logged-out session satisfied both of the old conditions and kept counting as active for up to the full 30-day lifetime, in exactly the screen a user opens to check whether logout worked.

The same miscount also kept those sessions **out of** the account activity log, which deliberately lists everything that is *not* an active login (`Api::V1::User::AccountActivityController#model_index`) -- so a logged-out session was simultaneously shown as active and missing from the history.

Reproduced against the dev DB: 6 of the 323 tokens the old logic reported as active were already soft-killed, spread over 3 users, and all 6 have an `AccountActivity` record that was being suppressed. The new logic reports 317.

## The change

`active_logins` now just delegates to doorkeeper-mongodb's own `not_expired` scope, which already encodes exactly this per document -- including `revoked_at`, and treating a nil `expires_in` as "never expires" per `Doorkeeper::Expirable#expired?`. No reason to rebuild that comparison by hand.

`expired_logins` gets the inverse as a new `expired` scope on the token model, so the pair stays a true partition of a user's unrevoked tokens. It has no callers today, but leaving one half on the old `created_at` arithmetic would hand the next person a subtly wrong sibling -- and after this change a soft-killed token would otherwise have fallen out of *both* sets.

### The part that is not cosmetic

Narrowing the scope would have silently removed a capability, so this PR keeps it explicitly. `#destroy` resolves through `MODEL`, `#index` through `MODEL_OVERVIEW`, and in `Api::V1::User::AccountLoginsController` both pointed at `active_logins`. A soft-killed token's `refresh_token` is still live on purpose, and deleting the record is the only way to revoke that leftover credential from a **different** device -- "forget account" only reaches the browser holding it. `MODEL` is therefore now deliberately wider than `MODEL_OVERVIEW`: unrevoked own tokens, still scoped to `current_user.oauth_tokens`, so nothing widens across users.

## What was verified

Both specs were confirmed **red against the pre-fix code first**, not just green after:

```
# reverting user.rb only
11 examples, 2 failures
  User#active_logins excludes a token soft-killed by logout, even though it was created just now
  User#expired_logins includes the token soft-killed by logout

# narrowing MODEL back to active_logins
5 examples, 1 failure
  the destroy scope (MODEL) still resolves a logged-out session, so its surviving refresh_token stays revocable
```

- The `$expr` query cross-checked against a pure-Ruby expiry computation over a 431-token mixed sample (8 dead, 423 alive) -- **identical id sets**, not merely identical counts. My first cross-check ran over the oldest 3000 tokens and matched at `0 == 0`; that was a check that could not fail, so it was redone against a sample that actually contains both cases.
- The scope survives the full index chain the base controller builds (`quickfilter` / `gridfilter` / `auto_includes` / `sorting` / `paginate`): selector keeps the `$or`, no soft-dead token leaks through.
- No nil `created_at` anywhere in the collection (such a document would fall out of both sets and break the partition), and `resource_owner_id` is indexed, so the `$expr` only ever evaluates over one user's tokens.
- A second `include IdentityManagementExtension` does not raise `Mongoid::Errors::ScopeOverwrite`, so adding a `scope` to a concern that an initializer pushes into a gem class is safe here.
- `Api::V1::User::AuthenticateOtpController` also names `active_logins` as `MODEL`, but `index`/`show`/`update`/`destroy` are undef'd there and `create` is overridden to use `current_token` -- and no `before_action` in the base controller evaluates `MODEL`, so the OTP login path is untouched. Checked, because narrowing this scope wrongly would have broken MFA login outright.
- `spec/models` + `spec/controllers`: 136 examples, 0 failures. No spec fixtures left behind in the shared dev DB.
- rubocop: **zero new offenses** in the three app files (163 / 34 / 2, unchanged against `origin/main`); the remaining spec-file cops are the same ones the already-merged specs in this repo carry.

The second commit came out of the local review pass rather than from a reviewer: nothing asserted that the *index* scope stays restricted to the requesting user. `not_expired` builds its condition with Mongoid's `.or`, the one combinator that folds the receiver's criteria into a branch instead of ANDing onto it. The `resource_owner_id` does still land above the `$or` today, so there is no leak -- but a gem upgrade reshaping that scope would turn this endpoint into a cross-user session listing with nothing going red. Pinned, and confirmed the guard fails when `active_logins` is pointed at an unscoped `not_expired`.

## Deliberately out of scope

- **Frontend follow-up, filed separately, not papered over here:** the sessions that are now correctly absent from the active list still hold a live `refresh_token`, and there is no UI to see or revoke them. `expired_logins` is the set to build that on.
- **Gap 1** (the remembered-account skip is effectively a 30-day MFA-satisfied credential in `localStorage`) -- a product decision, not a code fix.
- **Gap 2** (the refresh grant does not retire the old access token) -- real, separate change, touches the remembered-account logic.
- A hard-revoked token that is *also* past its expiry is in neither set. That is deliberate -- it is a revoked session, not an expired one -- and matches the previous behaviour, since `expired_logins` already filtered on `revoked_at: nil`.

## Risk

The behaviour change is user-visible in two places by design: the active-logins list gets shorter, and the account activity log gets longer by the same sessions. Both were verified against real data, but only on the dev DB -- production has a larger and older token population, and the only shape that would behave differently there is a document with a nil `created_at`, which does not exist in dev.

Not verified end-to-end through HTTP: these are scope-level specs following this repo's existing controller-spec convention (`spec/controllers/api/v1/user/tenants_controller_spec.rb`), so the serializer and route layers above `model_index` / `model_destroy` are covered by inspection only.

### DIFF
--- a/app/controllers/api/v1/user/account_logins_controller.rb
+++ b/app/controllers/api/v1/user/account_logins_controller.rb
@@ -2,10 +2,26 @@ class Api::V1::User::AccountLoginsController < Api::V1::JsonApiController
   API = :public
 
   MODEL_BASE = Doorkeeper::AccessToken
+
+  # #destroy resolves through MODEL, #index through MODEL_OVERVIEW (see
+  # Api::V1::JsonApiController#model_destroy / #model_index), and here MODEL is
+  # deliberately WIDER than what the list shows.
+  #
+  # A token soft-killed by logout (Api::V1::Doorkeeper::TokensController#revoke
+  # writes expires_in: -1 for the token_type_hint: 'access_token' branch) is no
+  # longer an active session and is correctly absent from #active_logins -- but its
+  # refresh_token deliberately survives logout, because that is what powers the
+  # remembered-account fast path on the login page. Deleting the record here is the
+  # only way to revoke that surviving refresh credential from a DIFFERENT device
+  # ("forget account" only reaches the browser holding it), so it has to stay
+  # addressable by id even though it is no longer listed (#2422). Scoped to
+  # current_user.oauth_tokens throughout, so this widens nothing across users.
   MODEL = -> {
+    current_user.oauth_tokens.where(revoked_at: nil).order(created_at: -1)
+  }
+  MODEL_OVERVIEW = -> {
     current_user.active_logins.order(created_at: -1)
   }
-  MODEL_OVERVIEW = MODEL
   SERIALIZER = AccountLoginSerializer
   OVERVIEW_SERIALIZER = AccountLoginSerializer
 
--- a/app/models/concerns/identity_management_extension.rb
+++ b/app/models/concerns/identity_management_extension.rb
@@ -28,6 +28,28 @@ module IdentityManagementExtension
     field :im_otp_provided, type: Mongoid::Boolean
     field :im_otp_tries, type: Integer, default: 0
 
+    # Inverse of doorkeeper-mongodb's own `not_expired` scope, restricted to tokens
+    # that were never hard-revoked: tokens whose own expiry window has closed. Used
+    # by User#expired_logins so that pair stays a true partition of a user's
+    # unrevoked tokens (#2422).
+    #
+    # expires_in is per-document, so the comparison needs $expr -- this mirrors the
+    # expression in doorkeeper-mongodb's `not_expired` (mixins/mongoid/
+    # access_token_mixin.rb) with the operator inverted. expires_in is in seconds
+    # while $add treats a plain number added to a date as milliseconds, hence *1000.
+    # A nil expires_in is excluded on purpose: Doorkeeper::Expirable#expired? treats
+    # nil as "never expires", so such a token is active, not expired.
+    scope :expired, -> {
+      where(revoked_at: nil, :expires_in.ne => nil).where(
+        '$expr' => {
+          '$lte' => [
+            { '$add' => ['$created_at', { '$multiply' => ['$expires_in', 1000] }] },
+            Time.now.utc
+          ]
+        }
+      )
+    }
+
     before_create :fill_computed
     before_save :fill_computed
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -877,16 +877,31 @@ class User < ApplicationDocument
     @access_group_ids
   end
 
+  # Sessions the user currently has: not hard-revoked, and their OWN expiry window
+  # still open. Both of these used to derive that window from created_at plus the
+  # GLOBALLY configured lifetime, ignoring the token's own expires_in -- so a token
+  # soft-killed on logout (Api::V1::Doorkeeper::TokensController#revoke writes
+  # expires_in: -1 for the token_type_hint: 'access_token' branch) kept counting as
+  # an active session until its created_at aged past that global lifetime: up to 30
+  # days after the user logged out, in the very screen someone opens to check that
+  # logout worked (#2422). It also kept those sessions out of the account activity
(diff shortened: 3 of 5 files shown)


SCjona (commented):
LGTM — no blocking findings.

Verified against the actual code, not just the patch:

- `not_expired` in doorkeeper-mongodb 5.5.1 (`access_token_mixin.rb:58-75`) builds `relation = where(revoked_at: nil)` and then `relation.where($expr …).or(relation.where(expires_in: nil))`. Because *both* branches are built from `relation`, the association's `resource_owner_id` ends up inside each $or branch, so `oauth_tokens.not_expired` stays per-user. The cross-user guard spec pins exactly the failure mode a gem upgrade could introduce here — good call.
- The new `expired` scope is a true complement: the gem uses `$gt` on `created_at + expires_in*1000`, this uses `$lte` on the same expression, with `expires_in: nil` excluded (never-expires ⇒ never expired) and `revoked_at: nil` on both sides. `expires_in: -1` lands in `expired`, as intended. It only uses `.where`, so it ANDs onto the receiver rather than folding it into a branch.
- No name collision: the gem's mixin declares only `not_expired`, and `Doorkeeper::Models::Expirable` contributes `expired?` as an instance method, so `scope :expired` in `IdentityManagementExtension` does not shadow anything.
- The widened `MODEL` is reachable from `#destroy` only — `show`/`create`/`update` are `undef_method`'d in `Api::V1::User::AccountLoginsController`, and `JsonApiController#model_show/#model_create/#model_update` are the only other consumers of `MODEL`. It stays under `current_user.oauth_tokens`, so nothing widens across users, and `AccountLoginSerializer` exposes no token material.
- `Api::V1::User::AuthenticateOtpController` names `active_logins` as `MODEL` but undef's `index`/`show`/`update`/`destroy` and overrides `create` to use `current_token`, so the MFA path is unaffected.
- The refresh grant is not affected either: `Api::V1::App::Doorkeeper::TokensController#create` issues a *new* token and stamps `expires_in: access_token_expires_in` on it, while the previous token keeps `expires_in: -1` and only has its `refresh_token` randomised — so a re-login via the remembered-account path shows up as a fresh active session, not a resurrected dead one.
- `Api::V1::User::AccountActivityController#model_index` is the only other caller; its `.not.where(token_id: current_user.active_logins.pluck(:id))` gets longer by exactly the sessions that leave the active list, which is the stated intent.

The nil-`created_at` document that would fall out of both sets is already called out in the PR description with the production caveat, so there is nothing to add.

SCjona (approved):

PR #283, opened 2026-08-20
Stop the resend throttle on password reset from leaking account existence

neongrau:
## Summary
Re-opens #2419 (email address enumeration, Cognisys pentest). The first fix (#278) closed the enumeration vector on the *first* request to the account-recovery endpoint. This one closes the same class of bug on a second axis, found and reproduced by the E2E suite maintainer (`SCjona`, reopening comment): `Api::V1::Devise::PasswordsController#create` answered identically for a registered/unregistered address on the first request, but a **second** request to the same address within the 5-minute resend window returned a distinct `400 reset_password_delayed` for a registered address, while an unregistered one kept getting the generic `200`. The throttle branch itself was the oracle — exactly the "or the account's configuration" case the pentest recommendation named explicitly.

## Fix
Keep the resend throttle (no duplicate mail sent), but make it silent: whether the address doesn't exist, exists and gets mailed, or exists and is throttled, the controller now always returns the same generic `200 {success: true, message: ...}`.

## Changes
- `app/controllers/api/v1/devise/passwords_controller.rb` — merge the throttle check into the mail-sending condition instead of returning early with a distinct response; added a `SECURITY / PENTEST NOTE`
- `config/locales/{en,de,nl,fr,ru}.yml` — remove the now-unused `devise.mailer.reset_password_delayed` key (confirmed via org-wide code search: no other backend or frontend reader)
- `spec/controllers/api/v1/devise/passwords_controller_spec.rb` — new regression spec (zero prior coverage): asserts identical `200` body across all three branches, asserts the throttled case never sends a second mail, and directly asserts byte-identical response bodies for a throttled registered user vs. an unregistered one

## Verified
- Ran the new spec against the pre-fix code by stashing the fix: 3/7 failures, reproducing exactly the reported oracle (`400` vs `200`, and a body diff naming the leaked `reset_password_delayed` error). Restored the fix: 7/7 pass, 11/11 across both recovery specs.
- Org-wide code search confirms `identity-management-frontend`'s only consumer of this endpoint (forgot-password flow) does not branch on the removed error code or status.

Fixes Samedis-care/samedis-care-issues#2419


neongrau:
Local review pass complete (10-angle high-effort pass, 0 findings). Verified: (1) the frontend's only POST consumer of this endpoint (`ForgotPassword.tsx`) shows a generic error dialog and doesn't branch on status/error code, so removing the distinct 400 doesn't break it; (2) the removed `devise.mailer.reset_password_delayed` locale key has no other reader anywhere in the org (checked via org-wide code search); (3) the new spec fails against the pre-fix code (verified by stashing the fix and re-running: 3/7 failures reproducing the exact reported oracle) and passes with the fix (7/7, 11/11 combined with the sibling recoveries spec).

neongrau:
pr review

SCjona:
AI review in progress (this may take a few minutes)...

SCjona (commented):
LGTM — no blocking findings.

Verified the fix in `app/controllers/api/v1/devise/passwords_controller.rb:16-30`: all three branches (unregistered, registered-and-mailed, registered-and-throttled) now fall through to the same `render_jsonapi_msg({ success: true, message: reset_password_msg }, 200)`. `render_jsonapi_msg` (`app/controllers/concerns/json_api.rb:68`) emits only `locale` + `msg`, so the bodies really are byte-identical — the spec's `expect(throttled_body).to eq(unregistered_body)` is deterministic, not flaky. The throttle itself is preserved (no second mail inside the 5-minute window), and `send_reset_password_instructions` still refreshes `reset_password_sent_at`, so the window keeps working. `reset_password_delayed` is gone from all five locale files and has no remaining reader in the repo. This matches the KB fix for M-11 ("Throttle-Message = Success-Message", `knowledge/IMB/Security Audit - Pentest Prep 2026-07.md:193-196`).

Non-blocking: `spec/openapi/internal/paths/api/v1/devise/passwords.yaml:50` still documents a `400 – Invalid or unknown email` response for `POST /api/v1/{app}/users/password`. After this change the action can only ever return 200, so the published contract now advertises an enumeration-shaped response that no longer exists. Worth dropping that response block in this PR.

Non-blocking (pre-existing, follow-up issue): the sibling endpoint from the same M-11 audit item, `app/controllers/api/v1/app/recoveries_controller.rb:13-31`, still answers 400 (`recovery_email_unset`) when the address is unknown or has no recovery email, versus 200 when it is registered *and* has one — so account existence is still partially enumerable there. Out of scope for this PR, but the audit's recommended fix ("Einheitliche neutrale 200-Antwort unabhängig von Existenz") is only half done until that one is aligned too.

neongrau:
Thanks for the review — both points addressed:

1. **OpenAPI drift (important, fixed)**: `spec/openapi/internal/paths/api/v1/devise/passwords.yaml` still documented a `400 – Invalid or unknown email` response for `POST`, which the action can no longer produce. Dropped that response block and reworded the `200` description to state it's the generic, always-returned response (2026-08-20T02c89da). Verified: the file still parses as YAML and the `put` action's own (unrelated, legitimate) `400` block is untouched.

2. **RecoveriesController still partially enumerable (correctly out of scope here)**: confirmed against the M-11 finding in `IMB/Security Audit - Pentest Prep 2026-07.md` — you're right that the KB recommendation was for a full always-200 response on both endpoints, and `RecoveriesController` only got the narrower fix from PR #278. Filed Samedis-care/samedis-care-issues#2562 to track it, and flagged in the issue that fully closing it would go beyond what PR #278's maintainer explicitly asked for (a revert to the pre-`72341e1` generic-error behavior, not a full always-200 rewrite) — so it needs a confirm before implementing, not a silent fix. Not resolving this PR since it doesn't touch `RecoveriesController`.

neongrau:
pr review

SCjona:
AI review posted.

SCjona (commented):
LGTM — no blocking findings.

Re-review of commit `02c89da` (only new change since my last pass: the OpenAPI spec).

Both non-blocking notes from the previous round are settled:

1. **OpenAPI drift — fixed.** `spec/openapi/internal/paths/api/v1/devise/passwords.yaml` no longer advertises the `400 – Invalid or unknown email` response that `create` can no longer produce, and the `200` description now documents it as the always-returned generic response. Checked that the `put` action's own `400` (line 117, `Invalid token or password`) is untouched — that one is still reachable via `render_jsonapi_error(..., 'reset_token_invalid', 400)` in `passwords_controller.rb:45`, so keeping it is correct. The edited block is well-formed: the `>-` folded scalar's continuation lines are indented under it and `content:` returns to the sibling indent, and the `#2419` in the description sits inside the block scalar, so it is literal text rather than a YAML comment.
2. **RecoveriesController — accepted as out of scope.** Agreed with the reasoning; tracking it in Samedis-care/samedis-care-issues#2562 with an explicit confirm-before-implementing note is the right call given PR #278's maintainer asked for the narrower revert, not a full always-200 rewrite.

The behavioural fix in `app/controllers/api/v1/devise/passwords_controller.rb:16-30` is unchanged from what I verified last round: all three branches still converge on the single generic `200`, the throttle still suppresses the duplicate mail, and `reset_password_delayed` has no remaining reader in the repo.

Posting as a comment rather than an approval because `main` is a protected branch and this review should not count toward its required approvals. Nothing of mine is blocking the PR — my previous review was also a comment, so there is no REQUEST_CHANGES to dismiss.

SCjona (approved):

